Issue #9, opened 2017-01-17
Set up CI server for OSX and Windows


mousetraps:
This will help us detect potential issues earlier and also make it more broadly apparent what the current status of the parser is.

mousetraps:
Just added a travis config file in #20, and AppVeyor is also enabled, but we still need to add an appveyor config file to test on Windows.

mousetraps:
And we also need to update the travis config file to add an OS X build.

PR #20, opened 2017-01-20
Add travis config for Linux

mousetraps:
Part of #9 

Framework validation tests, which take much longer, are run separately from the core test suites.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ env:
 
 os:
   - linux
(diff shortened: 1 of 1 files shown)


mousetraps:
Well, that's rather inconvenient... https://github.com/travis-ci/travis-ci/issues/2313

Temporarily removing OS X build until we address this issue, so we can merge this in, and have *something* running.

PR #27, opened 2017-01-20
Fix Node API test failures

### LINKED ISSUE #26: fix "api" testsuite failures

mousetraps:
A couple of references hadn't been updated after removing the Node::kind property

close #26


mousetraps:
Passing!
image

PR #31, opened 2017-01-20
add failing grammar tests to skipped.json

### LINKED ISSUE #25: Make "grammar" testsuite pass
Add any tests we don't intend to fix immediately to `skipped.json` so we can more easily detect regressions.

mousetraps:
close #25


mousetraps:
Tests are now all green! We should make a point to revisit everything in `skipped.json` and `skippedLexical.json`, but at least now we'll be able to easily detect regressions. We don't require validation tests to all pass, but those are also run continually on every change so we can always have a pulse on our current status through the grammar.

PR #34, opened 2017-01-21
Optimize memory usage of tree traversal

### LINKED ISSUE #23: Ensure tree traversal operations (e.g. Node::getDescendantNodesAndTokens) are passing by reference
Otherwise we end up unnecessarily doubling peak memory while iterating through nodes/tokens.

nikic:
Fixes #23 per last comment. I've also dropped some references while at it.

### DIFF
--- a/src/Node.php
+++ b/src/Node.php
@@ -155,9 +155,6 @@ class Node implements \JsonSerializable {
     public function getChildNodesAndTokens() : \Generator {
         foreach ($this->getChildNames() as $name) {
             $val = $this->$name;
-            if (\is_string($val)) {
-                continue;
-            }
             if (\is_array($val)) {
                 foreach ($val as $child) {
                     if ($child !== null) {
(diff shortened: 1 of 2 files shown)


src/Node.php
```diff
@@ -210,6 +214,33 @@ public function getChildTokens() {
...
     /**
+     * Gets array of declared child keys (cached).
+     *
+     * This is used as an optimization when iterating over nodes: For direct iteration
+     * PHP will create a properties hashtable on the object, thus doubling memory usage.
+     * We avoid this by iterating over just the keys instead.
+     *
+     * @return string[]
+     */
+    private function getChildKeys() {
```
mousetraps: You mentioned you had a getSubNodeNames function for php-parser. Should we consider making this method public rather than private? And if so, it would make sense to rename to `getChildNames` for consistency.

src/Node.php
```diff
@@ -210,6 +214,33 @@ public function getChildTokens() {
...
+        $class = \get_class($this);
+        if (!isset(self::$childKeys[$class])) {
+            $keys = [];
+            $reflectionClass = new \ReflectionClass($class);
+            foreach ($reflectionClass->getProperties() as $property) {
+                if ($property->name === "parent" || $property->name === "kind") {
```
mousetraps: I removed the "kind" property recently, so we can actually remove this check

nikic:
I've signed the CLA yesterday, but specified a different email than on my GitHub profile. Just tried again with the email on GH, let's see if it works.

msftgits:
Closing & reopening this issue to see if the updated CLA is picked up or not.

PR #38, opened 2017-01-22
Use "yield from" in descendant generators

nikic:
This required some adjustment to remove the dependence on specific keys, in particular checks for $idx=0 and use of iterator_to_array() with $use_keys=true.

See also discussion in #23.

### DIFF
--- a/src/Node.php
+++ b/src/Node.php
@@ -32,7 +32,7 @@ class Node implements \JsonSerializable {
      * @throws \Exception
      */
     public function getStart() : int {
-        $child = iterator_to_array($this->getChildNodesAndTokens())[0];
+        $child = $this->getChildNodesAndTokens()->current();
         if ($child instanceof Node) {
             return $child->getStart();
         } elseif ($child instanceof Token) {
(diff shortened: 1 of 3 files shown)


src/Node.php
```diff
@@ -374,7 +380,7 @@ public function & getFileContents() : string {
      * @return Node|null
      */
     public function getDescendantNodeAtPosition(int $pos) {
-        $descendants = iterator_to_array($this->getDescendantNodes());
+        $descendants = iterator_to_array($this->getDescendantNodes(), false);
```
mousetraps: Looks like iterator_to_array is unnecessarily getting used in a bunch of places, when iterating through elements of the generator should suffice. Just filed #39 to track this.

tests/ParserInvariantsTest.php
```diff
@@ -76,7 +76,7 @@ public function testNodesAllHaveAtLeastOneChild($filename, Node $sourceFileNode)
      * @dataProvider sourceFileNodeProvider
      */
     public function testEveryNodeSpanIsSumOfChildSpans($filename, Node $sourceFileNode) {
-        $treeElements = iterator_to_array($sourceFileNode->getDescendantNodesAndTokens());
+        $treeElements = iterator_to_array($sourceFileNode->getDescendantNodesAndTokens(), false);
         array_push($treeElements, $sourceFileNode);
```
mousetraps: replace stray instances of `array_push` with $array[] - my guess is that it mostly occurs in test code. Just filed an issue tracking here #40

PR #43, opened 2017-01-23
Improved Travis maintainability by reducing redundancy

Bilge:


mousetraps:
@Bilge - much better, thanks!

.travis.yml
```diff
@@ -23,7 +23,11 @@ before_script:
...
-  - if [[ "$VALIDATION" != "true" ]]; then phpunit --testsuite grammar; fi
-  - if [[ "$VALIDATION" != "true" ]]; then phpunit --testsuite api; fi
+  - |
+    if [[ $VALIDATION = true ]]; then
+      phpunit --testsuite validation
+    else
+      phpunit --testsuite invariants
+      phpunit --testsuite grammar
+      phpunit --testsuite api
+    fi
```
mousetraps: FYI - it turns out that tests weren't failing properly after this change. See https://github.com/Microsoft/tolerant-php-parser/pull/100/commits/9f502a93a9385ee627a14047d9c973026205a0ea for the fix. 

https://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin

Bilge: I suppose that's true: the `if` statement returns the exit code of the last command only. Some people chain commands together with `&&` to avoid this, but I think `set -e` is a cleaner approach. I believe Travis compiles every step into one big bash script, in which case you may prefer to specify `set -e` at the earliest stage rather than just the `script` step.

mousetraps: Good call, thanks! Moved it to `before_script` instead

PR #44, opened 2017-01-23
Remove scanNextToken args

tpunt:
Looks like this was missed from commit 0d9c0f2e9aaaab07d7410d1e41582a7da9fc07e8


tpunt:
Hmm, I just noticed that the `Lexer` class has been deprecated. It's probably not worth fixing this then?

mousetraps:
Hey, why not? It makes the project better, so thanks!

At this point, we're keeping the file around mainly for reference... perhaps it would be a good idea to add a comment and/or move it into `experiments` so that it doesn't confuse anyone who starts digging into the code.

PR #45, opened 2017-01-23
Update type comment for memberName property

tpunt:


mousetraps:
Looks great, thanks again! Really easy to miss this sort of stuff, so thanks for being thorough!

Issue #35, opened 2017-01-21
 Add a "vendor" top level namespace


remicollet:
From http://www.php-fig.org/psr/psr-4/

> The fully qualified class name MUST have a top-level namespace name, also known as a “vendor namespace”.

Ok, you can say "PhpParser" is the vendor... a way to circumvent PSR-4 (as the project SubNamespaceName is not mandatory.. one reason why I dislike this PSR). And PhpParser could be a bit too generic and may create conflict.

Al least, it conflicts with https://packagist.org/packages/nikic/php-parser

Proposal: use Microsoft\PhpParser

I understand this is only doable as part of a new major version, because of BC break.

mattacosta:
Duplicate of #22

mousetraps:
Thanks @remicollet! We'll take care of it - the API is caveated well-enough that it should be fine to take in immediately without worrying about a version change.

PR #55, opened 2017-01-25
Enhancement/node

decanus:
NodeKind can be gotten through the class name, Constants seem almost useless. We can validate through instanceof.

Also eliminates need for reflection, and turns Node into an abstract class.


tpunt:
There seems to be a lot of repetition with child classes implementing the `getNodeKindName()` method. It would be far better to simply declare that method in the abstract `Node` class, and use a little LSB:
```php
public function getNodeKindName() : string
{
    return static::class;
}
```

Or if the full namespace isn't desired, then do a little parsing:
```php
public function getNodeKindName() : string
{
    $parts = explode('\\', static::class);
    return $parts[count($parts) - 1];
}
```

Of course, if the Node kinds were going to see a lot of usage in the future (as seen with `TokenKind` in the parser), then it would probably be more optimal to keep the nodes identified by integers (because performance).

decanus:
@tpunt I was thinking of using your second solution, might be smart. But instead of integers can we not just test on the instance of the class?

tpunt:
@decanus Sure, but a simple integer comparison will be much more performant than an `instanceof` operation. Like I said though, it depends how much usage the node kind will see in the future.

decanus:
@tpunt Do you know if this method would even be efficient?

```php
public function getNodeKindName() : string
{
    $parts = explode('\\', static::class);
    return $parts[count($parts) - 1];
}
```

tpunt:
@decanus It's still going to be faster than what was there before, because it doesn't make use of reflection.

decanus:
@tpunt Yeah fair enough, I originally took out the reflection part due to concerns for speed.

mousetraps:
Sorry for the delayed response - was heads down on VS Code release activities last week.

Thanks for the PR, and some good discussion here :smiley:.

To give a bit of context, the original reason that `$kind` existed was because as @tpunt mentioned, I figured that integer comparisons would be faster than `instanceof`. After some experimenting, though, I realize I was actually wrong about this because while the comparison itself might be faster, it wasn't when combined with the property access. Plus, removing the `$kind` property resulted in a nice memory reduction.

This change, however, to remove `$kind`, was made right before the release, and I didn't have the time to clean up remaining references to getKind, so I chose to let them stand for the time being.

So instead of this change, which makes refactorings a bit of a burden because we need to remember to change the string every time we change the Node name, I like the idea @decanus mentioned in https://github.com/Microsoft/tolerant-php-parser/pull/55#issuecomment-275260083, and we can consider other methods if we find later that it's important to optimize serialization. 

@decanus, would you mind making these updates to the PR so we can merge in? Also, fyi #35 just went through, which causes some merge conflicts because the new namespace is `Microsoft\PhpParser`, rather than `PhpParser`

decanus:
@mousetraps I'll open a new pull request so I don't need to fix all the conflicts

decanus:
@mousetraps if you say that there is not really any difference with instanceof, can we not just stick to that for comparison? That would allow full deletion of the constants, and would reduce the size of the code base

mousetraps:
@decanus yep, that's right - the NodeKind constants can be removed at this point in favor of instanceof comparisons.

PR #56, opened 2017-01-25
Minor Updates

decanus:
::class is almost nicer than the string of that class, allows easier code browsing.

Single quotes are faster than double quotes, might be interesting to look into changing all of them in the codebase as a micro optimisation.

### DIFF
--- a/src/Utilities.php
+++ b/src/Utilities.php
@@ -55,7 +55,7 @@ class Utilities {
                             DiagnosticKind::Error,
                             "Unexpected modifier '" . $tokenKindToText[$modifier->kind] . "'",
                             $modifier->start,
-                            $modifier->getEnd() - $modifier->start
+                            $modifier->length
                         );
                     }
                 }


tpunt:
I wouldn't bother switching the quotation style. Single quotes aren't faster than double quotes, and these (questionable) micro optimisations will do nothing for the code base's overall performance. There are far better things to optimise here.

decanus:
@tpunt I thought it was once mentioned that single quotes are faster due to the fact that php doesn't first try to evaluate like it does with the double quotes. But wether its useful is to the performance is probable very questionable.

tpunt:
@decanus [It's a myth](http://nikic.github.io/2012/01/09/Disproving-the-Single-Quotes-Performance-Myth.html) :)

decanus:
@tpunt Thanks for that, interesting read 👍

decanus:
Added a little update to the Node class, seems like a minor cleanup.

rybakit:
@decanus Also worth reading: https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html

src/Node.php
```diff
@@ -347,22 +347,21 @@ public function getNodeKindName() : string {
...
+        }
+
+        $parent = $this->parent;
+        $siblings = $parent->getChildNodes();
+        foreach ($siblings as $idx=>$nextSibling) {
+            if (spl_object_hash($nextSibling) !== spl_object_hash($this)) {
+                continue;
             }
-        } elseif ($this instanceof Script) {
-            return $this->endOfFileToken->getEndPosition();
```
mousetraps: Looks like some logic got removed here. Script nodes are the uppermost-construct, and explicitly don't have a parent, so `getEndPosition` will not work on ScriptNodes now. 

Let's fix this, and then also add a couple test cases in NodeApiTest to make sure this doesn't break in the future.

src/Node.php
```diff
@@ -347,22 +347,21 @@ public function getNodeKindName() : string {
...
-            foreach ($siblings as $idx=>$nextSibling) {
-                if (spl_object_hash($nextSibling) === spl_object_hash($this)) {
-                    $siblings->next();
-                    $nextSibling = $siblings->current();
-                    return $nextSibling !== null
-                        ? $nextSibling->getFullStart()
-                        : $this->getRoot()->endOfFileToken->fullStart;
-                }
+        if (!isset($this->parent)) {
+            throw new \Exception("Unhandled node: " );
```
mousetraps: not related to your change in particular, but while you're here, might as well clean up the exception message to be more informative (e.g. perhaps include the node name after the colon?)

src/Node.php
```diff
@@ -347,22 +347,21 @@ public function getNodeKindName() : string {
...
+        }
+
+        $parent = $this->parent;
+        $siblings = $parent->getChildNodes();
```
mousetraps: not related to your change in particular, but reading through this made me realize that $parent isn't getting used anywhere else, so it makes sense to just remove that assignment, and use `$siblings = $this->parent->getChildNodes();`

src/Node.php
```diff
@@ -347,22 +347,21 @@ public function getNodeKindName() : string {
...
+        foreach ($siblings as $idx=>$nextSibling) {
+            if (spl_object_hash($nextSibling) !== spl_object_hash($this)) {
+                continue;
             }
-        } elseif ($this instanceof Script) {
-            return $this->endOfFileToken->getEndPosition();
+
+            $siblings->next();
+            $nextSibling = $siblings->current();
+            return $nextSibling !== null ? $nextSibling->getFullStart() : $this->getRoot()->endOfFileToken->fullStart;
```
mousetraps: more scannable to keep the ternary on a couple lines, as it was before.

src/Token.php
```diff
@@ -65,21 +65,21 @@ public static function getTokenKindNameFromValue($kindName) {
     function jsonSerialize() {
         $kindName = $this->getTokenKindNameFromValue($this->kind);
 
-        if (!isset($GLOBALS["SHORT_TOKEN_SERIALIZE"])) {
-            $GLOBALS["SHORT_TOKEN_SERIALIZE"] = false;
+        if (!isset($GLOBALS['SHORT_TOKEN_SERIALIZE'])) {
```
mousetraps: @decanus @tpunt @nikic - is there a particular convention around which quotation mark style is preferred?

nikic: As far as I'm aware, there is no particular convention on quoting style. Don't think it makes sense to change here, especially as other parts of the project predominantly use double quotes.

decanus: Yeah there isn't. I was still under the idea that single quotes are faster. But as they aren't this can be closed IMHO

src/Node.php
decanus: According to my ide Script does not exist anymore

mousetraps: Oh man, you're right! It should actually be SourceFileNode now. I must have missed this when I converged the node kind constants and the class names (in preparation for removing node kind constants altogether.)

PR #63, opened 2017-01-30
lint php files on travis builds

staabm:
this makes the travis build break when syntax errors are contained in *.php files.

to prevent errors like https://github.com/Microsoft/tolerant-php-parser/pull/62

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ cache:
 
 before_script:
   - phpenv config-rm xdebug.ini
+  - if find . -name "*.php" -path "./src/*" -path "./experiments/*" -path "./tools/*" -path "./syntax-visualizer/server/src/*" -exec php -l {} 2>&1 \; | grep "syntax error, unexpected"; then exit 1; fi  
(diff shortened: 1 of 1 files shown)


villfa:
@staabm :+1:  but it should be restricted to the _src_ directory only

staabm:
as /tools and /test etc. also cotains php sources I just included _all the php files_ to make sure we dont miss anything.

its a matter of a few seconds anyway, so no need to complicate things.

staabm:
the last missing error will be fixed by https://github.com/Microsoft/tolerant-php-parser/pull/65.

so after #65 landed this should get a green build.

mousetraps:
@staabm thanks for the contribution! Any chance you could also update to disclude `syntax-visualizer/example`? (also left a comment in #65 to explain why that's there)

For reference, the directories that should be included are:
* src
* tests (except subdirectories)
* experiments
* tools
* validation (except subdirectories)
* syntax-visualizer/server/src/*

staabm:
fixed per comments and rebased

Issue #12, opened 2017-01-17
Template string support


mousetraps:
Related to #11 - the current implementation is incomplete, and we need to think about a proper representation.

PR #64, opened 2017-01-30
TemplateExpression should extend Expression

villfa:
_TemplateExpression_ cannot work as it is since it tries to extend the class _Node_ which doesn't exist in the same namespace.
It should extend _Expression_ like all the others classes in the same namespace.


mousetraps:
FYI you might also be interested in #12 - the `parseTemplateExpression` code path shouldn't be hit at all now that we've moved to `PhpTokenizer.php`, so we need to revisit our representation of template strings.

PR #65, opened 2017-01-30
fixed syntax error in syntax-visualizer example file

staabm:
I guess this was not intentionally added, right?

### DIFF
--- a/syntax-visualizer/example/foo.php
+++ b/syntax-visualizer/example/foo.php
@@ -5,3 +5,4 @@ class {
     function A () {
 
     public function B () { }
+}


mousetraps:
@staabm Thx for the contribution! Actually, it was intentionally added. That folder is meant to demonstrate the syntax-visualizer tool in action, and that includes error tolerance strategies (see the [corresponding *.ast file](https://github.com/Microsoft/tolerant-php-parser/blob/master/syntax-visualizer/example/foo.php.ast#L29)). Sorry for the confusion!

PR #68, opened 2017-01-30
Add support for implementing interfaces in anonymous classes

tpunt:


tests/cases/parser/objectCreationExpression10.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+$a = new class() implements A {};
```
mousetraps: might be worth adding another test that includes both extends and implements

PR #72, opened 2017-02-01
NodeKind Refactoring

decanus:
As discussed, this removes the entire NodeKind class, usage of reflection and compares using instanceOf


src/Node.php
```diff
@@ -312,35 +310,11 @@ protected function getChildrenKvPairs() {
     }
 
     public function jsonSerialize() {
-        $kindName = self::getNodeKindNameFromValue($this->getKind());
+        $kindName = self::getNodeKindName();
```
nikic: Shouldn't this use `static::`?

src/Node.php
```diff
@@ -15,11 +15,9 @@ class Node implements \JsonSerializable {
...
     public $parent;
 
-    public function getKind() {
-        // TODO remove all references to getKind
-        $constants = (new \ReflectionClass("Microsoft\\PhpParser\\NodeKind"))->getConstants();
-        $nameParts = explode("\\",get_class($this));
-        return $constants[end($nameParts)];
+    public function getNodeKindName() : string {
+        $parts = explode('\\', static::class);
+        return $parts[count($parts) - 1];
```
nikic: An alternative implementation would be to use strrpos(), which avoid creating a temporary array.

src/Node.php
mousetraps: Good catch - it doesn't make much of a functional difference at the moment, but might help if we ever decide to override `getNodeKindName`

mousetraps:
FYI pushed some minor changes to the branch, as per @nikic's comments.

decanus:
@mousetraps Thanks for making the changes

PR #78, opened 2017-02-02
use strrpos() in getLineCharacterPositionFromPosition() as suggested in issue #71

### LINKED ISSUE #71: Use `strrpos` instead of `strpos` in `GetLineCharacterPositionFromPosition`
`strrpos` should be far more efficient than the current `strpos` + while loop. 
https://github.com/Microsoft/tolerant-php-parser/blob/master/src/Utilities.php#L77-L94

villfa:
fix #71

### DIFF
--- a/src/PositionUtilities.php
+++ b/src/PositionUtilities.php
@@ -26,26 +26,16 @@ class PositionUtilities {
      * @return LineCharacterPosition
      */
     public static function getLineCharacterPositionFromPosition($pos, $text) : LineCharacterPosition {
-        if ($pos >= \strlen($text)) {
-            $pos = \strlen($text);
+        $length = \strlen($text);
+        if ($pos >= $length) {
+            $pos = $length;
         } elseif ($pos < 0) {
             $pos = 0;
         }
 
-        $newlinePositions = [];
-        $newlinePos = -1;
-        while ($newlinePos = \strpos($text, "\n", $newlinePos + 1)) {
-            if ($newlinePos < $pos) {
-                $newlinePositions[] = $newlinePos;
-                continue;
-            }
-            break;
-        }
-
-        $lastNewline = \count($newlinePositions) - 1;
-        $char = $pos - ($lastNewline >= 0 ? $newlinePositions[$lastNewline] + 1 : 0);
-        $line = \count($newlinePositions);
-
+        $lastNewlinePos = \strrpos($text, "\n", -($length - $pos));
+        $char = $pos - ($lastNewlinePos === false ? 0 : $lastNewlinePos + 1);
+        $line = $pos > 0 ? \substr_count($text, "\n", 0, $pos) : 0;
         return new LineCharacterPosition($line, $char);
     }
 }
--- a/src/PositionUtilities.php
+++ b/src/PositionUtilities.php
@@ -26,14 +26,14 @@ class PositionUtilities {
      * @return LineCharacterPosition
      */
     public static function getLineCharacterPositionFromPosition($pos, $text) : LineCharacterPosition {
-        $length = \strlen($text);
-        if ($pos >= $length) {
-            $pos = $length;
+        $textLength = \strlen($text);
+        if ($pos >= $textLength) {
+            $pos = $textLength;
         } elseif ($pos < 0) {
             $pos = 0;
         }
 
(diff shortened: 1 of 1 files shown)


src/PositionUtilities.php
```diff
@@ -26,26 +26,16 @@ public static function getRangeFromPosition($pos, $length, $text) {
...
-            }
-            break;
-        }
-
-        $lastNewline = \count($newlinePositions) - 1;
-        $char = $pos - ($lastNewline >= 0 ? $newlinePositions[$lastNewline] + 1 : 0);
-        $line = \count($newlinePositions);
-
+        $lastNewlinePos = \strrpos($text, "\n", -($length - $pos));
+        $char = $pos - ($lastNewlinePos === false ? 0 : $lastNewlinePos + 1);
```
mousetraps: I don't think this works in the following case:
```php
$text = "hello\n there";
$pos = 3;
```

`$lastNewlinePos` would be `false` because there are no newlines before position 3, so `$char` would be `0` rather than the expected value of `3`.

The tests should have caught this, so in addition to updating the logic, could you add an extra assert?

src/PositionUtilities.php
```diff
@@ -26,26 +26,16 @@ public static function getRangeFromPosition($pos, $length, $text) {
      * @return LineCharacterPosition
      */
     public static function getLineCharacterPositionFromPosition($pos, $text) : LineCharacterPosition {
-        if ($pos >= \strlen($text)) {
-            $pos = \strlen($text);
+        $length = \strlen($text);
```
mousetraps: nit: `$textLength` rather than `$length` to be more descriptive

src/PositionUtilities.php
```diff
@@ -26,26 +26,16 @@ public static function getRangeFromPosition($pos, $length, $text) {
...
-            break;
-        }
-
-        $lastNewline = \count($newlinePositions) - 1;
-        $char = $pos - ($lastNewline >= 0 ? $newlinePositions[$lastNewline] + 1 : 0);
-        $line = \count($newlinePositions);
-
+        $lastNewlinePos = \strrpos($text, "\n", -($length - $pos));
+        $char = $pos - ($lastNewlinePos === false ? 0 : $lastNewlinePos + 1);
+        $line = $pos > 0 ? \substr_count($text, "\n", 0, $pos) : 0;
```
mousetraps: Wow - at first I thought the ternary was some micro-optimization to avoid the function call, but turns out that no, `substr_count` just has some non-intuitive behavior.

@nikic - According to the [docs](http://php.net/manual/en/function.substr-count.php), only a negative length should count from the end of haystack, but that's not what the function is [actually doing.]( https://github.com/php/php-src/blob/cbc416e8d34d7f653e9d31d450c01bb208aa7921/ext/standard/string.c#L5279-L5281) I just dug into it, and neither the [RFC](https://wiki.php.net/rfc/negative-string-offsets), nor the corresponding PR reference the expected behavior when zero-length is specified. Additionally, there aren't any tests for the zero-length case in `substr_count_basic.phpt`, so it strikes me as a simple oversight. What's PHP's stance on backwards compatibility for things like this? Docs bug or php bug?

src/PositionUtilities.php
mousetraps: Oh wait, mea culpa - nvm, apparently I fail at subtraction - this looks fine :smiley:

src/PositionUtilities.php
mousetraps: This was a small enough change, I went ahead and updated it myself, and will merge in as soon as the tests pass.

src/PositionUtilities.php
nikic: Looks like an implementation bug. The RFC says it has the "same behavior as substr", which does the right thing for length=0. As 7.1 is early in the release cycle and this is new functionality (in PHP 7.0 length=0 was not allowed), we can fix this without problems. Can you please file a bug on bugs.php.net?

nikic: Filed and fixed https://bugs.php.net/bug.php?id=74041.

mousetraps: Awesome! And thanks for filling me in on general process. I know exactly where to go next time :smiley:. Gotta say, I've been really happy with the RFC process - it's made tracking down some of the original reasoning behind decisions like this really pleasant.

PR #84, opened 2017-02-03
fix fatal error from Lexer

villfa:
For the given code:
```php
<?php
use Microsoft\PhpParser\Lexer;

$content = '<?php echo PHP_EOL;';
$lexer = new Lexer($content);
$tokens = $lexer->getTokensArray();
```

I got the following error:
```
PHP Fatal error:  Uncaught Error: Call to undefined method Microsoft\PhpParser\Token::getTextForToken() in src/Lexer.php:217
```

Since this class is not used anymore I'm wondering if it should even be kept in the code base.


tpunt:
The `Lexer` class (and what could be done with it) was mentioned [here](https://github.com/Microsoft/tolerant-php-parser/pull/44#issuecomment-274708303).

PR #93, opened 2017-02-07
remove dependence on index while iterating through test cases

mousetraps:
cc @tpunt


tpunt:
@mousetraps That's much better :)

PR #96, opened 2017-02-11
Updated parseNamespaceUseDeclaration to group clauses without braces

### LINKED ISSUE #87: Parser doesn't properly parse namespace-use-clause list
The below code gets parsed fine
```php
<?php
namespace A;
use A;
```

However, if you include another namespace-use-clause, it does not get parsed properly.
```php
<?php
namespace A;
use A, B;
```
image

runz0rd:
Fix #87

### DIFF
--- a/src/DiagnosticsProvider.php
+++ b/src/DiagnosticsProvider.php
@@ -65,6 +65,21 @@ class DiagnosticsProvider {
                     }
                 }
             }
+            elseif ($node instanceof Node\Statement\NamespaceUseDeclaration) {
+                if (count($node->useClauses->children) > 1) {
+                    var_dump($node->useClauses);
+                    foreach ($node->useClauses->children as $useClause) {
+                        if($useClause instanceof Node\NamespaceUseClause && !is_null($useClause->openBrace)) {
+                            yield new Diagnostic(
+                                DiagnosticKind::Error,
+                                "Semicolon expected.",
+                                $useClause->getEndPosition(),
+                                1
+                            );
+                        }
+                    }
+                }
+            }
         }
 
         foreach ($node->getChildNodesAndTokens() as $child) {
--- a/src/DiagnosticsProvider.php
+++ b/src/DiagnosticsProvider.php
@@ -67,7 +67,6 @@ class DiagnosticsProvider {
             }
             elseif ($node instanceof Node\Statement\NamespaceUseDeclaration) {
                 if (count($node->useClauses->children) > 1) {
-                    var_dump($node->useClauses);
                     foreach ($node->useClauses->children as $useClause) {
                         if($useClause instanceof Node\NamespaceUseClause && !is_null($useClause->openBrace)) {
                             yield new Diagnostic(
--- a/src/DiagnosticsProvider.php
+++ b/src/DiagnosticsProvider.php
@@ -66,12 +66,12 @@ class DiagnosticsProvider {
                 }
             }
             elseif ($node instanceof Node\Statement\NamespaceUseDeclaration) {
-                if (count($node->useClauses->children) > 1) {
+                if (\count($node->useClauses->children) > 1) {
                     foreach ($node->useClauses->children as $useClause) {
                         if($useClause instanceof Node\NamespaceUseClause && !is_null($useClause->openBrace)) {
                             yield new Diagnostic(
                                 DiagnosticKind::Error,
-                                "Semicolon expected.",
+                                "; expected.",
                                 $useClause->getEndPosition(),
                                 1
                             );
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -66,7 +66,6 @@ use Microsoft\PhpParser\Node\PropertyDeclaration;
 use Microsoft\PhpParser\Node\ReservedWord;
 use Microsoft\PhpParser\Node\StringLiteral;
 use Microsoft\PhpParser\Node\MethodDeclaration;
-use Microsoft\PhpParser\Node;
 use Microsoft\PhpParser\Node\Parameter;
 use Microsoft\PhpParser\Node\QualifiedName;
 use Microsoft\PhpParser\Node\RelativeSpecifier;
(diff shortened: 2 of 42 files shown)


mousetraps:
Awesome, thanks for the contribution! Closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience.

src/Parser.php
```diff
@@ -2559,39 +2559,47 @@ private function parseNamespaceUseDeclaration($parentNode) {
...
-                },
-                $namespaceUseDeclaration
-            );
-            $namespaceUseDeclaration->closeBrace = $this->eat(TokenKind::CloseBraceToken);
         }
         $namespaceUseDeclaration->semicolon = $this->eatSemicolonOrAbortStatement();
         return $namespaceUseDeclaration;
     }
 
+    private function parseNamespaceUseGroupClauses(NamespaceUseDeclaration $namespaceUseDeclaration) {
```
mousetraps: nit: `parseNamespaceUseGroupClauses` => `parseNamespaceUseGroupClauseList` and `$namespaceuseDeclaration` => `$parentNode` for consistency.

tests/cases/parser/namespaceUseDeclaration30.php
```diff
@@ -0,0 +1,3 @@
+<?php
+// TODO error
```
mousetraps: Looks like a skipped token is already getting properly produced here, so no need for this comment.

src/Parser.php
```diff
@@ -2559,39 +2559,47 @@ private function parseNamespaceUseDeclaration($parentNode) {
 
         $namespaceUseDeclaration->useKeyword = $this->eat(TokenKind::UseKeyword);
         $namespaceUseDeclaration->functionOrConst = $this->eatOptional(TokenKind::FunctionKeyword, TokenKind::ConstKeyword);
-        $namespaceUseDeclaration->namespaceName = $this->parseQualifiedName($namespaceUseDeclaration);
-        if (!$this->checkToken(TokenKind::OpenBraceToken)) {
+        if($this->lookahead(TokenKind::CommaToken)) {
```
runz0rd: Yea, lookahead for comma did seem like a poor design. So with these node revisions, we should end up with `NamespaceUseDeclaration` having `useClauses`, that always assumes theres a list (`NamespaceUseClauseList`), even with just one member? If so, whats the purpose of `groupClauses`  inside the new `NamespaceUseDeclaration`?

mousetraps: > So with these node revisions, we should end up with NamespaceUseDeclaration having useClauses, that always assumes theres a list (NamespaceUseClauseList), even with just one member? 

Yep! That said, it's not the best representation when it comes to memory usage. In the future, we may want to avoid defining properties that would be null, arrays with zero or one elements, etc. But that's an optimization that would apply to many Node types, not just this one.

> If so, whats the purpose of groupClauses inside the new NamespaceUseClause?

groupClauses refers to the delimited list that appears inside of the open braces and appends the namespaceName (`use A\B\{C\D,E\F}`), and useClauses refers a higher level (non-prepended) list of usings (`use A\B, C\D`)

Now, this is only one idea, and there are certainly other ways that you could represent this - the primary constraints being:
* higher level constructs are applied at the right scopes. For instance the function keyword in `use function A, B` should be "applied" to both child A and B (if you're at B, you should be able to inspect the parent node to see if it belongs in the function import table - you shouldn't have to inspect the first element in the list to know that.)
* all forms of usings defined in [the grammar](https://github.com/php/php-langspec/blob/master/spec/19-grammar.md#user-content-grammar-namespace-use-declaration) are represented, and in a way that makes it possible to add simple post-parse rules to produce any additional errors.
* all [invariants](https://github.com/Microsoft/tolerant-php-parser/blob/master/docs/Invariants.md) about the tree are satisfied.

Hope that helps, and feel free to prod further - there are a lot of nuances to this problem in particular, and I don't claim this is the perfect solution - so the more questions we ask, the better.

src/Parser.php
```diff
@@ -2556,42 +2556,62 @@ private function parseNamespaceDefinition($parentNode) {
...
+                return $this->isQualifiedNameStart($token) || $token->kind === TokenKind::FunctionKeyword || $token->kind === TokenKind::ConstKeyword;
+            },
+            function ($parentNode) {
+                $namespaceUseClause = new NamespaceUseClause();
+                $namespaceUseClause->parent = $parentNode;
+                $namespaceUseClause->namespaceName = $this->parseQualifiedName($namespaceUseClause);
+                if ($this->checkToken(TokenKind::AsKeyword)) {
+                    $namespaceUseClause->namespaceAliasingClause = $this->parseNamespaceAliasingClause($namespaceUseClause);
+                }
+                if ($this->checkToken(TokenKind::OpenBraceToken)) {
```
mousetraps: if clause => else clause

Let's say we have 
```php
use A\B as C

{}
```

Chances are the open brace is the start of a compound statement rather than a group use clause

src/DiagnosticsProvider.php
```diff
@@ -65,6 +65,20 @@ public static function getDiagnostics($node) : \Generator {
                     }
                 }
             }
+            elseif ($node instanceof Node\Statement\NamespaceUseDeclaration) {
+                if (count($node->useClauses->children) > 1) {
+                    foreach ($node->useClauses->children as $useClause) {
+                        if($useClause instanceof Node\NamespaceUseClause && !is_null($useClause->openBrace)) {
```
mousetraps: nit: backslash before built-ins functions names for consistency (PHP inlines some of it's internal functions, but requires a fully-specified name to take advantage of these optimizations from within namespaces - `is_null` in one of them, and `count` isn't, but we fully specify all internal functions so we can automatically take advantage of future optimizations)
also see https://github.com/Microsoft/tolerant-php-parser/commit/b875edcf7caa8410a4513dcf79b4add9e60831a2

src/DiagnosticsProvider.php
```diff
@@ -65,6 +65,20 @@ public static function getDiagnostics($node) : \Generator {
...
+                            yield new Diagnostic(
+                                DiagnosticKind::Error,
+                                "Semicolon expected.",
```
mousetraps: nit: `';' expected.` for consistency

Issue #103, opened 2017-02-13
Support parsing multiple exception types in catch clauses


mousetraps:
PHP 7.1 introduced support for catching multiple exception types:
https://wiki.php.net/rfc/multiple-catch

![image](https://cloud.githubusercontent.com/assets/762848/22870621/e999f0c2-f176-11e6-871a-07d976f89215.png)

We should handle this condition on the parser by parsing as a delimited list of qualified names, where the delimiter is the `TokenKind::BarToken`, and we should be able to reuse the logic from `parseQualifiedNameList`

nesl247:
Any update on this?

BPScott:
I've filed an issue with php-langspec as this is not currently documented and tested upstream: https://github.com/php/php-langspec/issues/216

TysonAndre:
My thoughts on solving this:

Add a new property to the tolerant-php-parser object representing the `catch{}` Node (i.e. CatchClause, parsed by parseCatchClause).

CatchClause->remainingQualifiedNameList would either be null or begin with `|` and end with `T1 [ | T2...]`. It would be parsed if the next token was `|`. It would be immediately after qualifiedName.

- In a future backwards incompatible release (1.0 if following semver?), rename CatchClause->qualifiedName to qualifiedNameList and remove remainingQualifiedNameList

jens1o:
@Zerquix18 Feel free to file a pr :)

felixfbecker:
@Zerquix18 Seriously, whining won't make it be fixed any faster. A PR will. This project is driven by community contributions.

PR #104, opened 2017-02-13
DelimitedList method for Node only elements 

runz0rd:
For #6

### DIFF
--- a/src/Node/DelimitedList.php
+++ b/src/Node/DelimitedList.php
@@ -31,4 +31,17 @@ class DelimitedList extends Node {
         }
         $this->children[] = $node;
     }
+
+    /**
+     * @return Node[]
+     */
+    public function getListElements() {
+        $listElements = [];
+        foreach ($this->children as $child) {
+            if($child instanceof Node) {
+                $listElements[] = $child;
+            }
+        }
+        return $listElements;
+    }
 }


src/Node/DelimitedList.php
```diff
@@ -31,4 +31,17 @@ public function addToken($node) {
...
     }
+
+    /**
+     * @return Node[]
+     */
+    public function getListElements() {
+        $listElements = [];
+        foreach ($this->children as $child) {
+            if($child instanceof Node) {
+                $listElements[] = $child;
```
runz0rd: Would it make sense for DelimitedList to have a $delimiter prop instead, to keep them more flexible? We already have DelimiterLists that do not have a constant delimiter and there may be more.

mousetraps: > We already have DelimiterLists that do not have a constant delimiter and there may be more.

We do? Sounds like a bug :smiley:. Could you elaborate?

In general, we keep away from holding any "descriptor" properties on Nodes because the memory cost of additional properties is 12-16 bytes each (depending on 32 bit or 64 bit), which very quickly adds up on large code bases.

runz0rd: In #103 we should parse `QualifiedNameList` with `TokenKind::BarToken` delimiter, but in most cases its `TokenKind::CommaToken`.
But its probable I just misunderstood your suggestion ("simply define the delimiter as a constant on DelimitedList") :smile:

runz0rd: I noticed a `// TODO consider allowing empty delimiter to be more tolerant` in `parseDelimitedList`, so if allow insertion of empty delimiters (if there are none), the `getListElements` can return even key values of children. 
If this does sound okay to you, how would you prefer to represent these empty delimiters?

roblourens:
It looks like `getElements` now fills this role. Please correct me if I'm wrong.

PR #106, opened 2017-02-14
#95 Add name resolution API

### LINKED ISSUE #95: Add API for `getFullyQualifiedName`
Named nodes should include API for `GetFullyQualifiedName`. 

Should follow rules defined here:
http://php.net/manual/en/language.namespaces.rules.php

mousetraps:
This isn't done yet, but it's working, so I wanted to get it out there for early feedback..
- adds name resolutions (`getResolvedName`) for all qualified names that are possible to be resolved and compile-time
- adds `getNamespacedName` method to all applicable node types.
- fix miscellaneous doc bugs
- rename ITokenStreamProvider => TokenStreamProviderInterface
- still needs comprehensive tests (lots of edge cases) + cleanup
- will depend on #96

close #95


src/Node.php
```diff
@@ -430,4 +434,137 @@ public function getDocCommentText() {
...
+        // The idea: Keep a single set of import tables cached based on a unique root node id, and invalidate
+        // cache whenever we attempt to resolve a qualified name with a different root node.
+        //
+        // In order to make this work, it will probably make sense to change the way we parse namespace definitions.
+        // https://github.com/Microsoft/tolerant-php-parser/issues/81
+        //
+        // Currently the namespace definition only includes a compound statement or semicolon token as one if it's children.
+        // Instead, we should move to a model where we parse future statements as a child rather than as a separate
+        // statement. This would enable us to retrieve all the information we would need to find the fully qualified
+        // name by simply traveling up the spine to find the first ancestor of type NamespaceDefinition.
```
nikic: Note that imports do not necessarily occur at the top of a namespace. As such, the same name may resolve differently depending on where it occurs inside a namespace. For example:

```
namespace A;
new X; // new \A\X;
use B\X;
new X; // new \B\X;

Comment (mousetraps) [inline src/Node.php]: Foiled again! :wink: Thanks, I know you mentioned that earlier, but I hadn't fully grokked the implications there. I'll try to handle that case, and then think through some better ways to represent this. It might also be that while retrieving the namespaced names is simple enough, it doesn't practically make as much sense to include name resolution in the parser API (at least not in language server contexts)

File: src/Node.php [hunk 2]
```diff
@@ -430,4 +434,137 @@ public function getDocCommentText() {
...
+                // Add the alias and resolved name to the corresponding namespace, function, or const import table.
+                // If the alias already exists, it will get replaced by the most recent using.
+                // TODO - worth throwing an error here instead?
+                if ($alias !== null) {
+                    if ($functionOrConst === null) {
+                        // namespaces are case-insensitive
+                        $alias = \strtolower($alias);
+                        $namespaceImportTable[$alias] = ResolvedName::buildName($namespaceNameParts, $contents);
+                    } elseif ($functionOrConst->kind === TokenKind::FunctionKeyword) {
+                        // functions are case-insensitive
```
mousetraps: @felixfbecker - fyi currently unhandled case in language server

felixfbecker: Yep, that's a `wontfix` for the LS. See https://github.com/felixfbecker/php-language-server/issues/69 for context

mousetraps: Good to know, thx for the pointer!

Issue #83, opened 2017-02-03
Add PHPCS for code style validation


decanus:
Just an idea to slowly start improving the code style

mousetraps:
A git pre-commit hook would be slick too. ✨

decanus:
We could think about adding travis

felixfbecker:
Yes please. It's pretty easy:

https://github.com/felixfbecker/php-language-server/blob/master/phpcs.xml.dist
https://github.com/felixfbecker/php-language-server/blob/master/.travis.yml#L18

PR #107, opened 2017-02-15
Parse catch multiple exceptions

runz0rd:
For #103

Theres still the issue of detecting class names that do not implement Throwable. Any idea where can this fit in?


src/Parser.php
```diff
@@ -2453,10 +2457,10 @@ private function parsePropertyDeclaration($parentNode, $modifiers) {
         return $propertyDeclaration;
     }
 
-    private function parseQualifiedNameList($parentNode) {
+    private function parseQualifiedNameList($parentNode, $delimiter = TokenKind::CommaToken) {
         return $this->parseDelimitedList(
             DelimitedList\QualifiedNameList::class,
-            TokenKind::CommaToken,
+            $delimiter,
```
mousetraps: Oh, I see what you meant in the delimited list PR - yeah, I don't think this is a super common case to handle, so I think it makes sense to special case this usage (or come up with some other idea that doesn't add additional memory usage.)

src/DiagnosticsProvider.php
```diff
@@ -79,6 +79,17 @@ public static function getDiagnostics($node) : \Generator {
                     }
                 }
             }
+            elseif ($node instanceof Node\CatchClause) {
+                $lastElement = \end($node->qualifiedNameList->children);
+                if($lastElement instanceof Token && $lastElement->kind == TokenKind::BarToken) {
```
mousetraps: nit: space between if and paren here and everywhere else. sorry, we don't have automatic code style checking yet #83

src/Parser.php
```diff
@@ -1863,14 +1863,18 @@ private function parseCatchClause($parentNode) {
         $catchClause->parent = $parentNode;
         $catchClause->catch = $this->eat(TokenKind::CatchKeyword);
         $catchClause->openParen = $this->eat(TokenKind::OpenParenToken);
-        $catchClause->qualifiedName = $this->parseQualifiedName($catchClause); // TODO generate missing token or error if null
+        $catchClause->qualifiedNameList = $this->parseQualifiedNameList($catchClause, TokenKind::BarToken);
+        if(\is_null($catchClause->qualifiedNameList)) {
+            $catchClause->qualifiedNameList = new MissingToken(TokenKind::QualifiedName, $this->getCurrentToken()->fullStart);
+        }
```
mousetraps: Nice!

src/DiagnosticsProvider.php
mousetraps: Also === instead of ==

src/Parser.php
```diff
@@ -1863,14 +1863,18 @@ private function parseCatchClause($parentNode) {
         $catchClause->parent = $parentNode;
         $catchClause->catch = $this->eat(TokenKind::CatchKeyword);
         $catchClause->openParen = $this->eat(TokenKind::OpenParenToken);
-        $catchClause->qualifiedName = $this->parseQualifiedName($catchClause); // TODO generate missing token or error if null
+        $catchClause->qualifiedNameList = $this->parseQualifiedNameList($catchClause, TokenKind::BarToken);
```
mousetraps: Let's also add a new test case to handle this scenario.

mousetraps:
> Theres still the issue of detecting class names that do not implement Throwable. Any idea where can this fit in?

Cool idea, and probably best left as an exercise to the consumer :wink:. For that, we'd need to resolve imports, build a symbol table, etc., which are more "language service-y" things to do. cc @felixfbecker

src/Parser.php
```diff
@@ -1863,14 +1863,18 @@ private function parseCatchClause($parentNode) {
...
+        if(\is_null($catchClause->qualifiedNameList)) {
```
felixfbecker: using `=== null` is a lot faster and does the same

felixfbecker:
I agree that checking the class extends `Throwable` is the domain of a language service :)

tests/cases/php-langspec/exception_handling/odds_and_ends.php
```diff
@@ -20,14 +20,22 @@ class X {}	// not derived from Exception
...
 	echo "L0: In catch-block int\n";
 }
+
+//catch (InvalidArgumentException | $e) { // Diagnosed; good
+//    echo "L0: In catch-block multiple Exception\n";
+//}
+
+catch (BadMethodCallException | InvalidArgumentException $e) {
+    echo "L0: In catch-block multiple Exception\n";
+}
```
runz0rd: Some of the grammar tests from this file were failing. This was intentionally added to test if a valid case would pass. Want this one removed?

mousetraps: Yep, let's revert this file but just make sure that there are handwritten test cases added for this scenario (which it looks like you already added). The spec tests are generated using [this script](https://github.com/Microsoft/tolerant-php-parser/blob/master/tests/generate_spec_tests.php) and will be overwritten whenever we take updates from [php-langspec](github.com/php/php-langspec). Original source [here](https://github.com/Microsoft/tolerant-php-parser/blob/master/php-langspec/tests/exception_handling/odds_and_ends.phpt).

roblourens:
Hey @runz0rd, this looks ok to me - sorry it fell through the cracks. If you're still up for it, I would accept it if you can get it merged with the latest stuff in master.

Also, where did the `--TEST--` and `--EXPECT--` come from in the test?

mousetraps:
> Also, where did the --TEST-- and --EXPECT-- come from in the test? I don't see that in other php-langspec tests. CI is failing, not sure whether that's the reason.

@roblourens those look like the raw php language spec test format. The files in `tests/cases/php-langspec` are autogenerated using the `tests/generate_spec_tests.php` script, and shouldn't be changed manually.

If upstream changes happen in php/php-langspec, they can be pulled in using git subtree. Then run `generate_spec_tests.php` to regenerate test cases for the parser.

roblourens:
Got it, thanks!

roblourens:
This will be handled in #244, and now would be a breaking change.

PR #118, opened 2017-04-25
Implement yield expressions - fix #3

### LINKED ISSUE #3: yield-expression support
Right now "yield" and "yield from" are being treated as skipped tokens - update parser to add proper support. 
https://github.com/php/php-langspec/blob/master/spec/19-grammar.md#user-content-grammar-yield-expression

roblourens:


roblourens:
Fixes #3

mousetraps:
🎉 🎉 🎉

PR #136, opened 2017-06-08
fix var annotations

jens1o:


src/Node/ElseClauseNode.php
```diff
@@ -14,6 +14,6 @@ class ElseClauseNode extends Node {
     public $elseKeyword;
     /** @var Token */
     public $colon;
-    /**@var StatementNode | StatementNode[] */
+    /** @var StatementNode | StatementNode[] */
```
felixfbecker: Are spaces between union types allowed? I thought a space separated the type from a description

jens1o: I think it's forbidden, but I don't want to break codestyle to be honest.

felixfbecker: Don't break forbidden code style?

jens1o: Yep, you got it. Until I get a message, I won't break that.

roblourens: It does appear to be wrong, at least it isn't parsed by phpDocumentor when it's space-separated. There are a ton of those in the parser, I'll fix them all at once. Also some in the language server.

PR #137, opened 2017-06-08
break when token was found

jens1o:
There was a little bottleneck, because once we found it, we iterated through others without any sense.

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -57,6 +57,7 @@ class Token implements \JsonSerializable {
         foreach ($constants as $name => $val) {
             if ($val == $kindName) {
                 $kindName = $name;
+                break;
             }
         }
         return $kindName;


src/Token.php
```diff
@@ -57,6 +57,7 @@ public static function getTokenKindNameFromValue($kindName) {
         foreach ($constants as $name => $val) {
             if ($val == $kindName) {
```
jens1o: we could consider doing it strictly, because then php doesn't need to convert it(because most of the time this IS false).

src/Token.php
```diff
@@ -57,6 +57,7 @@ public static function getTokenKindNameFromValue($kindName) {
...
                 $kindName = $name;
```
felixfbecker: Can't you `return $name` directly here?

felixfbecker:
I saw this too! 👍

src/Token.php
jens1o: That was my thought, but maybe in a later pr it will get some deeper parsing ;)

jens1o:
But this time I was faster 😛

src/Token.php
felixfbecker: Should be easy enough to change it then ;)

PR #139, opened 2017-06-09
Add NodeIterator

### LINKED ISSUE #130: Implement RecursiveIterator
PHP has native interfaces and classes for Iterators (`Generator` implements `Iterator`). Iterators can be traversed with `foreach` and composed easily, for example by filtering through [`CallbackFilterIterator`](http://de2.php.net/manual/en/class.callbackfilteriterator.php).

In particular, it has the [`RecursiveIterator`](http://de2.php.net/manual/en/class.recursiveiterator.php) interface, that adds two more methods to implement: `hasChildren()` and `getChildren()`.
This allows it to be traversed with a [`RecursiveIteratorIterator`](http://de2.php.net/manual/en/class.recursiveiteratoriterator.php), which has countless options for emitting all nodes or just leaves, using level order, etc. `ParentIterator` can be used to visit just nodes with child nodes, `RecursiveLimitIterator` to limit the nodes traversed, `AppendIterator` to concatenate Iterators, `RecursiveTreeIterator` can be used to draw an ASCII tree of an AST and `RecursiveCallbackFilterIterator` to filter nodes (including skipping children).

The Parser already uses Generators for traversal, which is great, but it only has limited options for filtering and skipping certain children through a callback: https://sourcegraph.com/github.com/Microsoft/tolerant-php-parser/-/blob/src/Node.php#L158-159
I think providing class-based implementations for Iterators would give greater flexibility.

felixfbecker:
Closes #130

### DIFF
--- a/docs/Traversal.md
+++ b/docs/Traversal.md
@@ -41,7 +41,7 @@ Naive example of traversing all nodes in the current scope:
 
 ```php
 // Find all nodes in the current scope
-$nodesInScopeReIt = new \RecursiveCallbackFilterIterator($node, function ($current, $key, Iterator $iterator) {
+$nodesInScopeReIt = new \RecursiveCallbackFilterIterator($node, function ($current, string $key, \RecursiveIterator $it) {
     // Don't traverse into function nodes, they form a different scope
     return !($current instanceof Node\Expression\FunctionDeclaration);
 });
@@ -55,7 +55,7 @@ Building on that example, to get all variables in that scope us a non-recursive 
 
 ```php
 // Filter out all variables
-$vars = new \CallbackFilterIterator($it, function ($current, $key, $iterator) {
+$vars = new \CallbackFilterIterator($it, function ($current, string $key, \Iterator $it) {
     return $current instanceof Node\Expression\Variable && $current->name instanceof Token;
 });
 
--- a/docs/Traversal.md
+++ b/docs/Traversal.md
@@ -42,7 +42,7 @@ Naive example of traversing all nodes in the current scope:
 ```php
 // Find all nodes in the current scope
 $nodesInScopeReIt = new \RecursiveCallbackFilterIterator($node, function ($current, $key, Iterator $iterator) {
-    // Don't traverse into function nodes, they form a differnt scope
+    // Don't traverse into function nodes, they form a different scope
     return !($current instanceof Node\Expression\FunctionDeclaration);
 });
 // Convert the RecursiveIterator to a flat Iterator
@@ -72,4 +72,4 @@ You can convert your iterator to a flat array with
 $arr = iterator_to_array($it, true);
 ```
 
-The `true` ensures that the array is indexed numerically and not by Iterator keys (otherwise Nodes later Nodes with the same key will override previous).
+The `true` ensures that the array is indexed numerically and not by Iterator keys (otherwise later Nodes with the same key will override previous Nodes).
--- a/docs/Traversal.md
+++ b/docs/Traversal.md
@@ -1,0 +1,75 @@
+
+# AST Traversal
+
+All Nodes implement the `IteratorAggregate` interface, which means their immediate children can be directly traversed with `foreach`:
+
+```php
+foreach ($node as $key => $child) {
+    var_dump($key)
+    var_dump($child);
+}
+```
+
+`$key` is set to the child name (e.g. `parameters`).
+Multiple child nodes may have the same key.
+
+The Iterator that is returned to `foreach` from `$node->getIterator()` implements the `RecursiveIterator` interface.
+To traverse all descendant nodes, you need to "flatten" it with PHP's built-in `RecursiveIteratorIterator`:
+
+```php
+$it = new \RecursiveIteratorIterator($node, \RecursiveIteratorIterator::SELF_FIRST);
+foreach ($it as $node) {
+    var_dump($node);
+}
+```
+
+The code above will walk all nodes and tokens depth-first.
+Passing `RecursiveIteratorIterator::CHILD_FIRST` would traverse breadth-first, while `RecursiveIteratorIterator::LEAVES_ONLY` (the default) would only traverse terminal Tokens. 
+
+## Exclude Tokens
+
+To exclude terminal Tokens and only traverse Nodes, use PHP's built-in `ParentIterator`:
+
+```php
+$nodes = new \ParentIterator(new \RecursiveIteratorIterator($node, \RecursiveIteratorIterator::SELF_FIRST));
+```
+
+## Skipping child traversal
+
+To skip traversal of certain Nodes, use PHP's `RecursiveCallbackIterator`.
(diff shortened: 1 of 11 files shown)


docs/Traversal.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Skipping child traversal
+
+To skip traversal of certain Nodes, use PHP's `RecursiveCallbackIterator`.
+Naive example of traversing all nodes in the current scope:
+
+```php
+// Find all nodes in the current scope
+$nodesInScopeReIt = new \RecursiveCallbackFilterIterator($node, function ($current, $key, Iterator $iterator) {
+    // Don't traverse into function nodes, they form a differnt scope
```
jens1o: typo

roblourens:
I like it but I'm curious what the perf impact is compared to a generator or a simple loop

felixfbecker:
Yes, would be worth to benchmark. My gut is telling me that it is faster than generators, cause they have been slow in PHP historically. The implementation maybe could also be optimised a bit more at the cost of uglier code (array index instead of ArrayIterator could be faster).

felixfbecker:
Fixed it, but the test is not complete yet. Any suggestions how the implementation could be optimized appreciated. Would also like to add an Iterator that walks upwards (following `parent`), but don't know a good name.

tests/api/NodeIteratorTest.php
```diff
@@ -5,15 +5,14 @@
...
-        const FILE_CONTENTS = <<<'PHP'
-<?php
-function foo($x) {
-    if ($x) {
-        var_dump($x);
-    }
-}
-foo();
-PHP;
+    const FILE_CONTENTS = '
```
jens1o: use `nowdoc` instead?

felixfbecker: what's the advantage?

roblourens:
I wrote a very basic perf test that parses and traverses the full syntax trees of every file in the Wordpress project. In that scenario, the Iterator solution is much slower. 

- A simple recursive function and loop over `CHILD_NAMES`: ~.88s
- Node's `getDescendantNodesAndTokens` generator-based helper: ~1.03s
- `RecursiveIteratorIterator`: ~1.88s

There seems to be a lot going on for each iteration. I hacked NodeIterator to replace the ChildNamesIterator with an index into the array, and I see it around 1.72s, so there's probably more optimization that can be done. I don't know whether I'd want to use it in the language server quite yet or call it the recommended way to traverse the tree. What do you think? It's too bad because perf aside, I would prefer this style.

jens1o:
Oh, the times are pretty poor....

roblourens:
Those times include parsing so the time diff spent iterating is more significant.

felixfbecker:
Too bad! Can you share the benchmark code? I will try to squeeze more performance out of it (that was my suspicion anyway). I prefer this style too because the returned Iterator maintains the semantics of being nested, not flat, but if it's that slower I wouldn't use it in the language server.

src/Node.php
```diff
@@ -150,6 +150,15 @@ public function getRoot() : Node {
     }
 
     /**
+     * Gets an Iterator to iterate all descendant nodes
+     *
+     * @return NodeIterator
+     */
+    public function getIterator() {
+        return new NodeIterator($this);
```
felixfbecker: woops, yes. didn't change it when I moved it to the namespace.

roblourens:
Sure, here: https://github.com/Microsoft/tolerant-php-parser/tree/iteratorPerfTest

Run with `php -d memory_limit=500M validation/iteratorPerf1.php` for 1, 2, 3. This is very quick and dirty and it's possible that I'm doing something dumb.

felixfbecker:
I tried everything I could to optimize, but the Generator API is twice as fast. I think the reason why is simply because the Generator API creates a single Generator object, while the RecursiveIterator will create another Iterator object for every node. But it should also be mentioned that "reimplementing" the traversal inline instead of using the the Generator API is 4x as fast, so the more ergonomic the API, the worse the performance.

```
✔ ~/git/tolerant-php-parser [iterator ↑·2|✚ 3…1]
23:18 $ php validation/iteratorPerf3.php
0102030405060708090100
Total nodes: 403197
MACHINE INFO
============
PHP int size: 8
PHP version: 7.1.0
OS: Darwin Felixs-MacBook-Pro.local 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64

PERF STATS
==========
Input Source Files (#): 101
Input Source Size (MB): 10.603403091431

Time Usage (seconds): 0.62322616577148
Memory Usage (MB): 0
✔ ~/git/tolerant-php-parser [iterator ↑·2|✚ 3…1]
23:18 $ php validation/iteratorPerf2.php
0102030405060708090100
Total nodes and tokens: 403197
MACHINE INFO
============
PHP int size: 8
PHP version: 7.1.0
OS: Darwin Felixs-MacBook-Pro.local 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64

PERF STATS
==========
Input Source Files (#): 101
Input Source Size (MB): 10.603403091431

Time Usage (seconds): 0.30183696746826
Memory Usage (MB): 0
✔ ~/git/tolerant-php-parser [iterator ↑·2|✚ 3…1]
23:19 $ php validation/iteratorPerf1.php
0102030405060708090100
Total nodes: 376330
MACHINE INFO
============
PHP int size: 8
PHP version: 7.1.0
OS: Darwin Felixs-MacBook-Pro.local 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64

PERF STATS
==========
Input Source Files (#): 101
Input Source Size (MB): 10.603403091431

Time Usage (seconds): 0.14891004562378
Memory Usage (MB): 0
✔ ~/git/tolerant-php-parser [iterator|✚ 2…1]
23:31 $
```

felixfbecker:
I think we should put it in perspective and benchmark real application code. I.e. benchmark the PHP language server indexing a project with one API vs the other. It may be twice as fast, but if the bottleneck is elsewhere and one API or the other is only indexing time of e.g. 10s vs 11s than I would still prefer the more advanced API if it makes the LS code more readable and DRY.

felixfbecker:
It also depends on the use case. Some users might just want to write a little script for a single large automatic code refactor and just need to write it quickly and then execute it once. We could document that there are essentially three ways to traverse ASTs that have trade-offs in performance vs API ergonomics. Users can then pick what they want.

roblourens:
I agree that we should try it out in a real-world situation. It may be that while visiting every node is slower, it's easier to traverse more efficiently or something like that. I'll revisit this once I'm through the current outstanding issues.

PR #144, opened 2017-06-10
Fix order of InlineHtml CHILD_NAMES

### LINKED ISSUE #141: InlineHtml child names switched
```php

class InlineHtml extends StatementNode {
    /** @var Token | null */
    public $scriptSectionEndTag;

    /** @var Token */
    public $text;

    /** @var Token | null */
    public $scriptSectionStartTag;

    const CHILD_NAMES = [
        'scriptSectionEndTag',
        'text',
        'scriptSectionStartTag'
    ];
}
```

Parsing this:
```php
<?php
        function a() {
            $a = 1;
            $b = 2;
        }
        a();
```
results in this AST:
image

note how `scriptSectionEndTag` is `null`, but `scriptSectionStartTag` is set.
that means `<?php` is clearly the `scriptSectionStartTag`, but when traversing, the end tag will be traversed first, then the text, then the start tag.

felixfbecker:
Closes #141

### DIFF
--- a/src/Node/Statement/InlineHtml.php
+++ b/src/Node/Statement/InlineHtml.php
@@ -11,17 +11,17 @@ use Microsoft\PhpParser\Token;
 
 class InlineHtml extends StatementNode {
     /** @var Token | null */
-    public $scriptSectionEndTag;
+    public $scriptSectionStartTag;
 
     /** @var Token */
     public $text;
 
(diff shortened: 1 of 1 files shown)


felixfbecker:
Hmm, I don't understand why the tests are failing. @roblourens ?

PR #156, opened 2017-06-18
Renamed getStart => getStartPosition for consistency.

dantleech:
Renamed `Node#getStart` to `Node#getStartPosition` to be consistent with `Node#getEndPosition`.

### DIFF
--- a/src/Node.php
+++ b/src/Node.php
@@ -29,10 +29,10 @@ abstract class Node implements \JsonSerializable {
      * @return int
      * @throws \Exception
      */
-    public function getStart() : int {
+    public function getStartPosition() : int {
         $child = $this->getChildNodesAndTokens()->current();
         if ($child instanceof Node) {
-            return $child->getStart();
+            return $child->getStartPosition();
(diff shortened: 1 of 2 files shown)


src/Node.php
```diff
@@ -354,7 +354,7 @@ public function jsonSerialize() {
     }
 
     /**
-     * Get the end index of a Node.
+     * Get the end position of a Node.
```
felixfbecker: index (=offset) is more precise than position. Position can also mean (line, column).

dantleech: This method uses the term `Position`. Maybe  we should change the API to `get[Start|End]Offset`?

roblourens: I've thought about this, 'position' is not ideal because the language server protocol uses it for line/col and the parser also has `LineCharacterPosition`. There are a ton of references to 'position' so it would be a pain to change, but either index or offset would be better.

roblourens:
Cleaning up PRs - I'm going to close this for the reasons in the comment above, but I still want to make these names consistent. I filed #166 to come up with better names.

PR #157, opened 2017-06-19
Fix use of count() on a string value

andwur:
This is no longer allowed in PHP 7.2, and is typically unwanted.
See https://wiki.php.net/rfc/counting_non_countables


roblourens:
Thanks for the PR!

PR #158, opened 2017-07-08
Allow mulitiple edits at the same offset.

dantleech:
Currently an Assertion error is thrown when text edits are made in the wrong order AND when multiple edits are made in the same place. Which I believe is a mistake as this works as expected until I run it in an environment with assertions turned on:

```php
$source = '<?php';
$textEdits[] = new TextEdit(0, 0, 'Hello');
$textEdits[] = new TextEdit(0, 0, 'World');
echo TextEdit::applyEdits($textEdits, $source);
```

Expected:

```
HelloWorld<?php
```

Actual: Assertion error is thrown.

If assertions are disabled (`zend.assertions = -1`) this works as expected.

In addtion I changed the assertion to an `OutOfBounds` exception due to the inconsitency of the behavior of `assert`.

### DIFF
--- a/src/TextEdit.php
+++ b/src/TextEdit.php
@@ -36,10 +36,14 @@ class TextEdit {
         $prevEditStart = PHP_INT_MAX;
         for ($i = \count($edits) - 1; $i >= 0; $i--) {
             $edit = $edits[$i];
-            \assert(
-                $prevEditStart > $edit->start && $prevEditStart > $edit->start + $edit->length,
-                "Supplied TextEdit[] must not overlap, and be in increasing start position order."
-            );
+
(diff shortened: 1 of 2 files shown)


roblourens:
Why would you expect `HelloWorld` and not `WorldHello`? I would think that Hello would be inserted at 0, then World inserted at 0. I think this is why that limitation is there in the first place - it leads to confusing situations.

dantleech:
I don't know if its good to have the limitation or not - I suppose it can be worked around by building up all of the edits at that offset into a single edit.

I do think that an exception would be better than assert in this case (as an exception will always be thrown whereas `assert` is env. specific).

roblourens:
I don't know why one check is an assert, and the other is an exception. I think it would make sense to make the assert an exception as well.

Issue #159, opened 2017-07-09
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to iterator_to_array() must implement interface Traversable, array given


Petah:
Get an error trying to run the example:

    PHP Fatal error:  Uncaught TypeError: Argument 1 passed to iterator_to_array() must implement interface Traversable, array given

on this line:

    var_dump(iterator_to_array($errors));

roblourens:
Fixed in #160

PR #160, opened 2017-07-09
Fix error

Petah:
Fix error PHP Fatal error:  Uncaught TypeError: Argument 1 passed to iterator_to_array() must implement interface Traversable, array given

https://github.com/Microsoft/tolerant-php-parser/issues/159


roblourens:
Thanks for the PR

Issue #162, opened 2017-07-19
Implement a `FunctionLike` interface


roblourens:
See https://github.com/felixfbecker/php-language-server/pull/438#issuecomment-316351727 and `isFunctionLike` in php-language-server's `ParserHelpers.php`

jens1o:
Am I wrong or is this a quite easy task?

I just add a new interface and map them to the points how the parserhelpers recognizes them?

roblourens:
Yes it would be quite easy.

jens1o:
I'm going to do this in the next minutes.

roblourens:
Ok, I started a stopwatch 😁

jens1o:
Quality before speed 😛

PR #167, opened 2017-08-07
Add Functionlike interface

jens1o:
Closes Microsoft/tolerant-php-parser#162

It feels strange to create a file with the namespace beginning with `Microsoft`...


roblourens:
Thanks!

Issue #11, opened 2017-01-17
Refresh string literal representation


mousetraps:
The representation of string literals is still a work-in-progress, and the implementation became even more inconsistent when we moved away from the handspun lexer towards PhpTokenizer. We should put some thought into how best to represent strings and template strings.

PythooonUser:
Hey,

here in #11 and in #168 you talk about that strings and template strings are still not implemented perfectly. Can you iterate on what exactly is still missing or not working?

I had a look at the parse output of e.g.
```php
<?php "Hello $world";
```

Which gives
```json
{
    "SourceFileNode": {
        "statementList": [
            [...]
            {
                "ExpressionStatement": {
                    "expression": {
                        "StringLiteral": {
                            "startQuote": {
                                "kind": "DoubleQuoteToken",
                                "fullStart": 6,
                                "start": 6,
                                "length": 1
                            },
                            "children": [
                                {
                                    "kind": "EncapsedAndWhitespace",
                                    "fullStart": 7,
                                    "start": 7,
                                    "length": 6
                                },
                                {
                                    "Variable": {
                                        "dollar": null,
                                        "name": {
                                            "kind": "VariableName",
                                            "fullStart": 13,
                                            "start": 13,
                                            "length": 6
                                        }
                                    }
                                }
                            ],
                            "endQuote": {
                                "kind": "DoubleQuoteToken",
                                "fullStart": 19,
                                "start": 19,
                                "length": 1
                            }
                        }
                    },
                    "semicolon": {
                        "kind": "SemicolonToken",
                        "fullStart": 20,
                        "start": 20,
                        "length": 1
                    }
                }
            }
        ]
        [...]
    }
}
```

And it returns an instance of `StringLiteral` which is essentially a non-delimited list of whitespace/characters and expression nodes. That sounds good to me and seems to be flexible enough.

Have you considered also making escape characters like `\n` its own token appearing as a children node in a string? Could that be helpful? I know that this distinction is made in language grammars (syntax highlighting) at least.

Cheers

roblourens:
Honestly I'm not sure what else we had in mind. Would have to look back at old notes.

> Have you considered also making escape characters like \n its own token appearing as a children node in a string?

Probably would not parse it at that granularity unless there was a good reason for it.

PythooonUser:
Yes, I see. The current implementation seems to be in a good state :)

roblourens:
[...]

PR #168, opened 2017-08-10
Parse template strings correctly

### LINKED ISSUE #12: Template string support

roblourens:
Fixes #12

Removes the old template string implementation, adds a new one.

There's probably more that could be done to make the string representation more sophisticated, per #11, but I'm not sure what that should be. This is simple and works quite well.


src/Parser.php
```diff
@@ -1025,6 +997,71 @@ private function parseStringLiteralExpression2($parentNode) {
...
+        return null;
+    }
+
+    private function parseTemplateStringSubscriptExpression($postfixExpression) : SubscriptExpression {
+        $subscriptExpression = new SubscriptExpression();
+        $subscriptExpression->parent = $postfixExpression->parent;
+        $postfixExpression->parent = $subscriptExpression;
+
+        $subscriptExpression->postfixExpression = $postfixExpression;
+        $subscriptExpression->openBracketOrBrace = $this->eat(TokenKind::OpenBracketToken); // Only [] syntax is supported, not {}
```
jens1o: What do you mean with the comment?

`"echo {$myFancyArray[5]}"` is valid too, and will output the sixth element of the array called `myFancyArray`,

src/TokenKind.php
```diff
@@ -156,14 +156,6 @@ class TokenKind {
...
     const InvalidBinaryLiteral = 308;
     const StringLiteralToken = 309;
-    const UnterminatedStringLiteralToken = 310;
-
-    const TemplateStringStart = 311;
-    const TemplateStringMiddle = 312;
-    const TemplateStringEnd = 313;
-    const NoSubstitutionTemplateLiteral = 314;
-    const UnterminatedNoSubstitutionTemplateLiteral = 315;
-    const UnterminatedTemplateStringEnd = 316;
```
jens1o: bracking changes here?

src/Node/Expression/Variable.php
```diff
@@ -24,7 +24,7 @@ class Variable extends Expression {
     public function getName() {
         if (
             $this->name instanceof Token &&
-            $name = substr($this->name->getText($this->getFileContents()), 1)
+            $name = ltrim($this->name->getText($this->getFileContents()), '$')
```
felixfbecker: So this trims all `$` from the left - what about variable variables? e.g. `$$var`

src/Parser.php
roblourens: I mean, `"$a[0]"` works, but `"$a{0}"` doesn't.

src/TokenKind.php
roblourens: These were never really used.

src/Node/Expression/Variable.php
roblourens: `getName` wouldn't work on `$$var`, because of the check above - `name` is a `Variable` node here. You would have to use `$var->name->getName()`.

I made this change for `"${x}"`. `x` is a variable that doesn't have a dollar sign, so I only want to remove a dollar sign if it exists.

src/Parser.php
jens1o: Ah, okay. Now I got that.

PR #174, opened 2017-09-13
Normalize spaces in union types in comments ("A | B" -> "A|B")

TysonAndre:
Mentioned in discussion of
https://github.com/Microsoft/tolerant-php-parser/pull/136/files 
(Couldn't find the code style guidelines referred to, if any. Is that just referring to the current state?)

Many IDEs and tools don't expect whitespace, and don't handle this.
So, they may act as if only the first type.

- https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types
  is a proposal, and is vague about how whitespace is handled
  outside `<>`
- https://github.com/phpDocumentor/phpDocumentor2/blob/develop/docs/references/phpdoc/tags/var.rst
  is vague
- https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/DocBlock/Tags/Var_.php
  doesn't expect any whitespace after `@var`

### DIFF
--- a/src/DiagnosticsProvider.php
+++ b/src/DiagnosticsProvider.php
@@ -9,14 +9,14 @@ namespace Microsoft\PhpParser;
 use Microsoft\PhpParser\Node;
 
 class DiagnosticsProvider {
-    /**
-     * Traverses AST to generate diagnostics.
-     * @param \Microsoft\PhpParser\Node $node
-     * @return \Generator | Diagnostic[]
-     */
 
     private static $tokenKindToText;
 
+    /**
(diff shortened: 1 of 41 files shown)


roblourens:
Thanks! I don't think the test files supposed to be changed?

TysonAndre:
Reverted the test files. I'm not sure when those changed, must have modified them somehow or checked out a stale version by accident.

After this was approved, `@param` and `@return` were also modified, and DiagnosticsProvider.php had phpdoc added and reordered.

PR #177, opened 2017-09-15
Fix incorrect PHPdoc for src/Parser.php, nit on property assignment

TysonAndre:
Clarify phpdoc of parseDelimitedList

Don't assign a DelimitedList to node->nameParts,
it's declared as an array.
(Has no side effects, this is overwritten afterwards unless nameParts was null)

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -345,7 +345,6 @@ class Parser {
 
     /**
      * Aborts parsing list when one of the parent contexts understands something
-     * @param ParseContext $context
      * @return bool
      */
     private function isCurrentTokenValidInEnclosingContexts() {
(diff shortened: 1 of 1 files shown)


src/Parser.php
```diff
@@ -1218,7 +1234,7 @@ function ($parentNode) {
                 return null;
```
roblourens: Previously it would return null if `$node->nameParts` was null, ie if `parseDelimitedList` returned null. I think this check should be for `$nameParts` now? Otherwise I agree with not using $node->nameParts to hold a temporary value.

TysonAndre: Yes, it should. Missed that.

roblourens:
Thanks!

PR #183, opened 2017-09-19
Remove an unnecessary reference from parameter.

TysonAndre:
The parameter isn't being modified, so having this is unexpected,
and passing params by reference is slower than passing by value.

The reference was part of the initial commit for src/Token.php


roblourens:
Thanks!

PR #184, opened 2017-09-20
Stop including non-php files in validation tests.

TysonAndre:
Noticed test failures for pgsql.phpunit.xml in the drupal project's test
files. (contains the substring ".php").
That also included `foo.php.gz`

Also, do the matching on the path name, not on the object to string
conversion.

### DIFF
--- a/tests/ParserFrameworkValidationTests.php
+++ b/tests/ParserFrameworkValidationTests.php
@@ -18,9 +18,11 @@ class ParserFrameworkValidationTests extends TestCase {
             $iterator = new RecursiveDirectoryIterator(__DIR__ . "/../validation/frameworks/" . $frameworkName);
 
             foreach (new RecursiveIteratorIterator($iterator) as $file) {
-                if (strpos($file, ".php") !== false) {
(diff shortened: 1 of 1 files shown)


tests/ParserFrameworkValidationTests.php
```diff
@@ -18,9 +18,11 @@ public function frameworkErrorProvider() {
             $iterator = new RecursiveDirectoryIterator(__DIR__ . "/../validation/frameworks/" . $frameworkName);
 
             foreach (new RecursiveIteratorIterator($iterator) as $file) {
-                if (strpos($file, ".php") !== false) {
+                $pathName = $file->getPathname();
+                if (preg_match('@\.php$@', $pathName) > 0) {
```
roblourens: Could you use `/` as the regex delimiter instead?

roblourens:
Thanks!

tests/ParserFrameworkValidationTests.php
TysonAndre: Done

PR #187, opened 2017-09-20
Speed up token value to name lookup

TysonAndre:
Grammar test suite sped up from 1.2 seconds to 1.1 seconds.

This went from O(n) (loop over (around half of) 183 constants) to O(1) (hash lookup)

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -59,15 +59,25 @@ class Token implements \JsonSerializable {
         return $this->fullStart + $this->length;
     }
 
-    public static function getTokenKindNameFromValue($kindName) {
-        $constants = (new \ReflectionClass("Microsoft\\PhpParser\\TokenKind"))->getConstants();
-        foreach ($constants as $name => $val) {
-            if ($val == $kindName) {
-                $kindName = $name;
-                break;
-            }
+    /**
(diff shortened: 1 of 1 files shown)


src/Token.php
```diff
@@ -59,15 +59,22 @@ public function getEndPosition() {
...
+            $mapToKind = \array_flip($constants);
         }
-        return $kindName;
+        return $mapToKind;
+    }
+
+    /** @return int (Or string, if $kindName wasn't found) */
+    public static function getTokenKindNameFromValue($kindName) {
+        $mapToKind = self::getTokenKindNameFromValueMap();
+        return $mapToKind[$kindName] ?? $kindName;
```
roblourens: The variable names are confusing here (and they were like that before). $kindName the parameter is really the value? And I think it should return "" or null if it isn't found.

TysonAndre: Was in a rush in the morning, noticed that $kindName was wrong but didn't fix it (Threw the rest of the documentation it was based off of off).

This appears to be used in generating debug data from tests, and I don't plan on changing the function behavior in this PR.

roblourens:
Sure. Thanks!

src/Token.php
roblourens: Sure. Thanks!

PR #196, opened 2017-09-26
Handle parenthesized breakout level

### LINKED ISSUE #195: continue(1) and break(1) should not be syntax errors
continue

The optional numeric constant for `break` and `continue` statements can be enclosed in parentheses which is erroneously reported as a syntax error:

`';' expected.`

See here: [http://sandbox.onlinephpfunctions.com/code/19bee6f4f7dc93a6a146950fd9877a438c131897](http://sandbox.onlinephpfunctions.com/code/19bee6f4f7dc93a6a146950fd9877a438c131897)

MadHed:
Fixes #195

`parseBreakOrContinueStatement` now parses `breakoutLevel` as an `Expression`.
The additional check that the breakout level must be a positive integer has been moved into `DiagnosticProvider`.


src/DiagnosticsProvider.php
```diff
@@ -84,6 +84,45 @@ public static function checkDiagnostics($node) {
...
+                if (
+                    $breakoutLevel instanceof Node\Expression\NumericLiteral
+                    && \in_array($breakoutLevel->children->kind, [
+                        TokenKind::BinaryLiteralToken,
+                        TokenKind::DecimalLiteralToken,
+                        TokenKind::HexadecimalLiteralToken,
+                        TokenKind::OctalLiteralToken,
+                        TokenKind::IntegerLiteralToken
+                    ])
+                    && \intval($breakoutLevel->getText()) > 0
```
felixfbecker: why not just cast to an int?

MadHed: I don't think that works for anything but decimal numbers

roblourens:
Thanks! We really need tests for DiagnosticsProvider...

MadHed:
Yup, that's what I figured writing this.
Maybe similar to the spec tests, we could compare them to a known serialized diagnostics output.

PR #198, opened 2017-09-27
Add diagnostics test

MadHed:
Added diagnostic tests to the test suite. Works the same as the tree tests.
This includes the .diag files for the current state of the parser.

Additionally I fixed a bug in the DiagnosticsProvider when a use declaration with an empty use clause list is checked.


roblourens:
Thanks, this is awesome! I think the test failures are just what was fixed by the other PR, right?

MadHed:
Yes, I see you just merged the other PR so I can rebase this branch on the new master. Just a moment :)

MadHed:
Ok, checks do pass now!

Please note that I have only added checks for the ParserGrammar tests, the lexer and validation tests do not do diagnostic checking. But I think this covers plenty of cases for now.

PR #199, opened 2017-10-03
Exclude more non-essential files

vinkla:
This pull request [exclude even more non-essential](https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production/) files when installing this package in production.

Happy Hacktoberfest!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,9 +2,18 @@
 /tests/*.php text eol=lf
 /tests/cases/**/*.php text eol=lf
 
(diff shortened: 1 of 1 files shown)


roblourens:
Thanks! Why aren't the individual files `export-ignore`?

vinkla:
> Thanks! Why aren't the individual files export-ignore?

Good catch! I've pushed an update.

PR #203, opened 2017-10-30
README: make require use of __DIR__

TomasVotruba:
When we now absolute path, it's better to use it

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ with the Abstract Syntax Tree (AST) via a friendly API.
 ```php
 <?php
 // Autoload required classes
-require "vendor/autoload.php";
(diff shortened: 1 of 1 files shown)


roblourens:
Why is that better? Shouldn't the examples be as simple as possible?

felixfbecker:
@roblourens otherwise the example doesn't work if the cwd doesn't match the directory of the file

TomasVotruba:
@roblourens Same way object code is better than static. It's more code, but more error proof. Moreover in long-lasting projects.

roblourens:
Thanks, I also fixed this in docs/GettingStarted.md

TomasVotruba:
Thanks! I didn't look there, thank you.

PR #207, opened 2017-11-11
nit: Remove duplicate array keys

TysonAndre:
Detected via static analysis

When there are duplicate array keys, the last entry takes precedence
(e.g. the removed T_EXIT line had the wrong value DieKeyword)


felixfbecker:
How can we get that static analysis into the language server? :)

TysonAndre:
There's an experimental Phan editor plugin, but it only works on unix/linux right now due to a `pcntl` requirement. https://github.com/tysonandre/vscode-php-phan#examples 

https://github.com/TysonAndre/tolerant-php-parser/tree/master/.phan is an example config on the tolerant-php-parser

PR #211, opened 2017-11-21
Optimize getDiagnostics for speed

TysonAndre:
Add a simple benchmarking script

Extreme optimization, sacrificing code style in some areas.

- This is meant for discussion, since the project is
  in the optimization phase. Avoiding nested generators did help performance.

This commit reduces the time needed to getDiagnostics
from 0.013 to 0.007 seconds. (On php 7.2.0RC5 with opcache enabled for CLI. Sample file was src/Token.php)

```
Average time to generate diagnostics (before): 0.013...
Average time to generate diagnostics (after):  0.006921
time to parse (Without diagnostics):           0.036374
```

(E.g. getDiagnostics previously took 36% of the time compared to generating an
AST, and now it takes 19% of that time.)

### DIFF
--- a/src/DiagnosticsProvider.php
+++ b/src/DiagnosticsProvider.php
@@ -13,122 +13,78 @@ class DiagnosticsProvider {
     private static $tokenKindToText;
 
     /**
+     * @param int $kind (must be a valid token kind)
+     * @return string
+     */
+    public static function getTextForTokenKind($kind) {
+        if (!isset(self::$tokenKindToText)) {
+            self::initTokenKindToText();
+        }
+        return self::$tokenKindToText[$kind];
+    }
+
+    /**
+     * @return string[]
+     */
+    private static function initTokenKindToText() {
+        self::$tokenKindToText = \array_flip(\array_merge(
+            TokenStringMaps::OPERATORS_AND_PUNCTUATORS,
+            TokenStringMaps::KEYWORDS,
+            TokenStringMaps::RESERVED_WORDS
+        ));
+    }
+
+    /**
      * Returns the diagnostic for $node, or null.
-     * @param \Microsoft\PhpParser\Node $node
+     * @param \Microsoft\PhpParser\Node|\Microsoft\PhpParser\Token $node
      * @return Diagnostic|null
      */
     public static function checkDiagnostics($node) {
-        if (!isset(self::$tokenKindToText)) {
-            self::$tokenKindToText = \array_flip(\array_merge(
-                TokenStringMaps::OPERATORS_AND_PUNCTUATORS,
-                TokenStringMaps::KEYWORDS,
-                TokenStringMaps::RESERVED_WORDS
-            ));
+        if ($node instanceof Token) {
+            if (\get_class($node) === Token::class) {
+                return null;
+            }
+            return self::checkDiagnosticForUnexpectedToken($node);
         }
 
-        if ($node instanceof SkippedToken) {
+        if ($node instanceof Node) {
+            return $node->getDiagnosticForNode();
+        }
+        return null;
+    }
+
+    /**
+     * @param Token $token
+     * @return Diagnostic|null
+     */
+    private static function checkDiagnosticForUnexpectedToken($token) {
+        if (!isset(self::$tokenKindToText)) {
+            self::initTokenKindToText();
+        }
+        if ($token instanceof SkippedToken) {
             // TODO - consider also attaching parse context information to skipped tokens
             // this would allow us to provide more helpful error messages that inform users what to do
             // about the problem rather than simply pointing out the mistake.
             return new Diagnostic(
                 DiagnosticKind::Error,
                 "Unexpected '" .
-                (self::$tokenKindToText[$node->kind]
-                    ?? Token::getTokenKindNameFromValue($node->kind)) .
+                (self::$tokenKindToText[$token->kind]
+                    ?? Token::getTokenKindNameFromValue($token->kind)) .
                 "'",
-                $node->start,
-                $node->getEndPosition() - $node->start
+                $token->start,
+                $token->getEndPosition() - $token->start
             );
-        } elseif ($node instanceof MissingToken) {
+        } elseif ($token instanceof MissingToken) {
             return new Diagnostic(
                 DiagnosticKind::Error,
                 "'" .
-                (self::$tokenKindToText[$node->kind]
-                    ?? Token::getTokenKindNameFromValue($node->kind)) .
+                (self::$tokenKindToText[$token->kind]
+                    ?? Token::getTokenKindNameFromValue($token->kind)) .
                 "' expected.",
-                $node->start,
-                $node->getEndPosition() - $node->start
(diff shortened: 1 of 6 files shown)


src/DiagnosticsProvider.php
```diff
@@ -12,125 +12,79 @@ class DiagnosticsProvider {
...
+            self::initTokenKindToText();
+        }
+        return self::$tokenKindToText[$kind];
+    }
+
+    /**
+     * @return string[]
+     */
+    private static function initTokenKindToText() {
+        self::$tokenKindToText = \array_flip(\array_merge(
```
jens1o: Could someone explain to me what the advantage of  `array_flip(array_merge([...]))` is?

felixfbecker: Creating a hashmap that can be looked up in O(1)?

jens1o: oh, that's cool. Nice trick, will use that when I need it :)

src/DiagnosticsProvider.php
```diff
@@ -12,125 +12,79 @@ class DiagnosticsProvider {
 
     private static $tokenKindToText;
 
+    /**
+     * @param int $kind (must be a valid token kind)
+     * @return string
+     */
+    public static function getTextForTokenKind($kind) {
+        if (!isset(self::$tokenKindToText)) {
```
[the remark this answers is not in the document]
roblourens: Yeah, I would be ok with initializing it after the class declaration.

src/Node.php
```diff
@@ -159,17 +159,54 @@ public function getDescendantNodesAndTokens(callable $shouldDescendIntoChildrenF
...
                 }
             } elseif ($child instanceof Token) {
                 yield $child;
             }
         }
     }
 
+    /**
+     * Iterate over all descendant Nodes and Tokens, calling $callback.
+     * This can often be faster than getDescendantNodesAndTokens if you just need to call something and don't be a generator.
```
roblourens: It's not quite clear when to use one or the other. Is it as simple as "convenient" vs "faster"? Should this be documented in the public API?

TysonAndre:
By public API, do you mean the markdown file with method descriptions?

Also, I should have written "don't need"

roblourens:
Yeah - I don't want to confuse people by documenting two methods which do basically the same thing. But, if this is significantly faster than the other one, then it makes sense for it to be documented.

Btw thanks for tackling this!

roblourens:
Sorry that this was left behind. I don't think the new method should be documented as public API at the moment. There are several methods that return generators, and we need to be consistent - we would need to decide, do we stick to generators officially, or provide alternates for all of them, or give up on generators entirely? 

If you remove that and fix the merge conflict, I'll merge this.

felixfbecker:
I wonder what the real-project perf impact of this is. In general I think the Generator/Iterator pattern is better than the visitor pattern.

@TysonAndre would you be willing to make a PR for the language server to use this and check the impact?

TysonAndre:
I'm not familiar with the benchmarks used for felixfbecker's php-language-server. https://github.com/felixfbecker/php-language-server/search?utf8=%E2%9C%93&q=benchmark&type= doesn't reveal anything.

My own use case was similar to the benchmark I included in experiments/ (Parse the AST, then immediately call getDiagnostics.

The summary of that was mentioned in the PR description: getDiagnostics previously took 36% of the time compared to generating an
AST, and now it takes 19% of that time.)

jens1o:
https://github.com/felixfbecker/php-language-server/blob/master/Performance.php

TysonAndre:
31.6 vs 31.4 seconds for php-language-server (Excluding the checks of tolerant-php-parser), there was a lot of variance, though.
I'd guess that parsing the AST is cheaper than everything else that's being done in php-language-server (That was the case for my application).


Using the same Benchmark.php with/without the updated composer.json and `composer.phar update` : https://github.com/felixfbecker/php-language-server/compare/master...TysonAndre:use-optimized-getDiagnostics?expand=1

With the change from this PR (second of two consecutive runs)

```
time php -d opcache.enable_cli=1 Performance.php                                                             
0
1000
2000
3000
4000
5000
6000
------------------------------
Time [drupal]: 17.724359035492
0
------------------------------
Time [wordpress]: 1.3051180839539
0
------------------------------
Time [php-language-server]: 0.28266787528992
0
------------------------------
Time [math-php]: 0.61766982078552
0
1000
2000
3000
------------------------------
Time [symfony]: 8.395693063736
0
------------------------------
Time [codeigniter]: 0.61100316047668
0
------------------------------
Time [cakephp]: 2.2087850570679
php -d opcache.enable_cli=1 Performance.php  31.04s user 0.37s system 99% cpu 31.408 total
```



Without the change (second of two consecutive runs):

```
» time php -d opcache.enable_cli=1 Performance.php
0
1000
2000
3000
4000
5000
6000
------------------------------
Time [drupal]: 17.855414867401
0
------------------------------
Time [wordpress]: 1.3178341388702
0
------------------------------
Time [php-language-server]: 0.28475403785706
0
------------------------------
Time [math-php]: 0.62388896942139
0
1000
2000
3000
------------------------------
Time [symfony]: 8.4373149871826
0
------------------------------
Time [codeigniter]: 0.61493396759033
0
------------------------------
Time [cakephp]: 2.2180669307709
php -d opcache.enable_cli=1 Performance.php  31.23s user 0.38s system 99% cpu 31.614 total
```

jens1o:
> I'd guess that parsing the AST is cheaper than everything else that's being done in php-language-server (That was the case for my application).

You may be interested in looking at https://github.com/felixfbecker/php-language-server/pull/451

roblourens:
The TreeAnalyzer iterates the tree on its own and calls checkDiagnostics, instead of getDiagnostics, so I guess this wouldn't impact the language server's perf, is that right?

roblourens:
Well getDiagnostics is called by the language server when a document is opened or changed: https://github.com/felixfbecker/php-language-server/blob/master/src/Server/TextDocument.php#L138 so it would help there. But I don't think it will show an impact in that perf test.

roblourens:
Never mind, that's not the same `getDiagnostics`. `DiagnosticsProvider::getDiagnostics` is actually not called by the language server. The language server has the TreeAnalyzer which is already using a loop + recursion to walk the tree. https://github.com/roblourens/php-language-server/blob/master/src/TreeAnalyzer.php#L57

The changes to `checkDiagnostics` are good with or without the extra method to walk descendent nodes. If that other method was part of the parser's public API, then the language server would use it there, but I still don't want to support that. And it doesn't make sense for the language server to use getDiagnostics since it also has to walk the tree to collect definitions and references. 

Anyway, I think it still makes sense to merge this, knowing that it doesn't really impact php-language-server's perf, but does help anyone else calling `getDiagnostics` a good amount (based on the original numbers).

Does this summary seem correct?

TysonAndre:
The summary seems correct.

The new method on Node isn't documented as part of the public API. I could add `@internal` if you felt it was necessary.

I assume there's nothing else that needs to be added?

PR #212, opened 2017-11-23
Efficiently look up lines/columns of many nodes

### LINKED ISSUE #179: Add built in way to efficiently look up line numbers for a large number of nodes/tokens?
This may be useful in the next phase (Making this easy to consume).
Currently, I only see one method, which counts the number of newlines up to that byte in the file. 
If an application determined line numbers for a majority of nodes, it would be slow (O(n*m), n being number of nodes, m being file size)

Example use cases:

- Language server which reports line number in error message, completion, etc.(?)
- Converting to other ast formats (https://github.com/Microsoft/tolerant-php-parser/issues/113)

Example API:

```
class LineNumberLocator {
    // Example strategies: eager (E.g. walk the tree and build a map from offsets to line numbers), or support binary search (E.g. in an array mapping line numbers to offsets), or the slow search.
    // Or cursor (assume the next requested node/token will be next to the previously requested one).
    public static function create(string $file_contents, Node $sourceFileNode, int $strategy = ...) : LineNumberLocator;

    public function getStartLine($node) : int;

    // can determine from getStartLine if an array mapping lines to offsets is built
    public function getStartColumn($node) : int;

    public function getEndLine($node) : int;

    public function getEndColumn($node) : int;
}
```

Other considerations: 

- Will this API's design be affected by incremental rebuilding?

TysonAndre:
Fixes #179, adds FilePositionMap

This assumes that requested offsets will be close to previous requests,
which should be true if this is used while walking a tree, etc.

Future tasks if the basic design is solid:

- copy license to new files
- Document this in the public API
- Deprecate PositionUtilities or make those APIs use FilePositionMap?
  (E.g. cache FilePositionMap in a static variable,
   and use $text is === as previous request)

### DIFF
--- a/src/FilePositionMap.php
+++ b/src/FilePositionMap.php
@@ -1,4 +1,8 @@
-<?php declare(strict_types=1);
+<?php
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
 
 namespace Microsoft\PhpParser;
 
@@ -9,11 +13,8 @@ namespace Microsoft\PhpParser;
  * Other designs that weren't chosen:
  * - Precomputing all of the start/end offsets when initializing was slower - Some offsets weren't needed, and walking the tree was slower.
  * - Caching line numbers for previously requested offsets wasn't really necessary, since offsets are usually close together and weren't requested repeatedly.
- *
- * Future optimizations:
  */
-class FilePositionMap
-{
+class FilePositionMap {
     /** @var string the full file contents */
     private $fileContents;
 
@@ -26,8 +27,7 @@ class FilePositionMap
     /** @var int the line number for $this->currentOffset (updated whenever currentOffset is updated) */
     private $lineForCurrentOffset;
 
-    public function __construct(string $file_contents)
-    {
+    public function __construct(string $file_contents) {
         $this->fileContents = $file_contents;
         $this->fileContentsLength = \strlen($file_contents);
         $this->currentOffset = 0;
@@ -40,24 +40,21 @@ class FilePositionMap
      * if https://github.com/Microsoft/tolerant-php-parser/issues/166 is fixed,
      * (i.e. if there is a consistent way to get the start offset)
      */
-    public function getNodeStartLine(Node $node) : int
-    {
+    public function getNodeStartLine(Node $node) : int {
         return $this->getLineNumberForOffset($node->getStart());
     }
 
     /**
      * @param Node $node the node to get the start line for.
      */
-    public function getTokenStartLine(Token $token) : int
-    {
+    public function getTokenStartLine(Token $token) : int {
         return $this->getLineNumberForOffset($token->start);
     }
 
     /**
      * @param Node|Token $node
      */
-    public function getStartLine($node) : int
-    {
+    public function getStartLine($node) : int {
         if ($node instanceof Token) {
             $offset = $node->start;
         } else {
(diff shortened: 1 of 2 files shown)


src/FilePositionMap.php
```diff
@@ -0,0 +1,146 @@
...
+            $offset = $node->getStart();
+        }
+        return $this->getLineNumberForOffset($offset);
+    }
+
+    /**
+     * @param Node|Token $node
+     * Similar to getStartLine but includes the column
+     */
+    public function getStartLineCharacterPositionForOffset($node) : LineCharacterPosition
```
roblourens: Should this be `ForOffset` or just `getStartLineCharacterPosition`?

src/FilePositionMap.php
```diff
@@ -0,0 +1,146 @@
...
+            return 1;
+        } elseif ($offset > $length) {
+            $offset = $length;
+        }
+        // Postcondition: offset >= 1, ($lastNewlinePos < $offset)
+        // If there was no previous newline, lastNewlinePos = 0
+
+        // Start strrpos check from the character before the current character,
+        // in case the current character is a newline.
+        $lastNewlinePos = \strrpos($this->file_contents, "\n", -$length + $offset - 1);
```
roblourens: When getting line and column, which is probably the most common case, we already know the position of the last newline. Wouldn't it be much faster to reuse that?

src/FilePositionMap.php
```diff
@@ -0,0 +1,146 @@
...
+ * Other designs that weren't chosen:
+ * - Precomputing all of the start/end offsets when initializing was slower - Some offsets weren't needed, and walking the tree was slower.
+ * - Caching line numbers for previously requested offsets wasn't really necessary, since offsets are usually close together and weren't requested repeatedly.
+ *
+ * Future optimizations:
+ */
+class FilePositionMap
+{
+    /** @var string the full file contents */
+    private $file_contents;
```
felixfbecker: Please use camelCase

src/FilePositionMap.php
TysonAndre: Just getStartLineCharacterPosition, will fix later.

src/FilePositionMap.php
TysonAndre: I'd rather not focus on micro-optimizing right now, this is simpler and fairly fast.

I was using substr_count, which doesn't tell me **where** the last newline was. It just tells me how many newlines there were.

I'd expect the native string functions to be faster than doing a for loop in PHP.
I haven't yet tried repeatedly calling `stripos(contents, "\n")`. I imagine the other approach could be faster if the number of lines skipped over was small.

src/FilePositionMap.php
TysonAndre: Will change

src/FilePositionMap.php
roblourens: That's fine

src/FilePositionMap.php
TysonAndre: Done for all properties, and the test continues to pass.

src/FilePositionMap.php
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * FilePositionMap can be used to get the line number for a large number of nodes (starting from 1).
+ * It works the most efficiently when the requested node is close to the previously requested node.
+ *
+ * Other designs that weren't chosen:
+ * - Precomputing all of the start/end offsets when initializing was slower - Some offsets weren't needed, and walking the tree was slower.
+ * - Caching line numbers for previously requested offsets wasn't really necessary, since offsets are usually close together and weren't requested repeatedly.
+ *
+ * Future optimizations:
```
roblourens: Empty "Future optimizations"?

roblourens:
Please add the license block to the new files, then I'll merge it

src/FilePositionMap.php
TysonAndre: Those became optimizations that I tried (but didn't actually improve performance).

Forgot to remove the section.

jens1o:
@roblourens It's added :)

PR #215, opened 2017-12-13
Use Null Coalesce Operator

carusogabriel:


tests/cases/php-langspec/interfaces/vector.php
```diff
@@ -43,20 +43,13 @@ public function offsetGet($offset)
...
 		echo "Inside " . __METHOD__ . " with offset >$offset<\n";
 
-		if (isset($this->elements[$offset]))
-		{
-			return $this->elements[$offset];
-		}
-		else
-		{
-			return NULL;
-		}
```
felixfbecker: These files look like they come from an external source, so please don't reformat them.

roblourens:
Thanks!

PR #217, opened 2017-12-13
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertInstanceOf` instead of `instanceof` operator;
- `assertNotFalse` instead of comparison with `false` keyword.

### DIFF
--- a/tests/api/getResolvedName.php
+++ b/tests/api/getResolvedName.php
@@ -82,7 +82,7 @@ class GetResolvedNameTest extends TestCase {
      */
     public function testGetResolvedName($contents, $expectedName) {
         $node = $this->getNodeAtPosition($contents);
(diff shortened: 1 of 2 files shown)


tests/api/getResolvedName.php
```diff
@@ -82,7 +82,7 @@ public function dataProvider() {
      */
     public function testGetResolvedName($contents, $expectedName) {
         $node = $this->getNodeAtPosition($contents);
-        $this->assertTrue($node instanceof Node\QualifiedName, "Node is not QualifiedName: " . get_class($node));
+        $this->assertInstanceOf(Node\QualifiedName::class, $node, "Node is not QualifiedName: " . get_class($node));
```
felixfbecker: I assume the custom message is not needed anymore?

carusogabriel: Gonna remove it soo.

PR #218, opened 2017-12-29
Tokenizer takes in consideration multibyte string lengths.

emilioastarita:
Hi!
I'm working in a port of tolerant-php-parser to golang (http://github.com/emilioastarita/gphp)

I'm using the output of your parser as validation of mine. One of the problem is that current php tokenizer is using `\strlen` to count string chars so it fails to count correctly when the string has multibyte chars. 

This PR uses `\mb_strlen` during tokenization to fix that.


roblourens:
Thanks for the PR! Some comments in the [docs](http://php.net/manual/en/function.mb-strlen.php) warn that `mb_strlen` can be much slower than `strlen`. Did you run the perf tests?

Do you know whether any other instances of `strlen` or other string methods need to be fixed?

emilioastarita:
Yes, is slower using `\mb_strlen`. But anyway I'm thinking better about this issue and I think that php tokenizer is right using `\strlen` at least if "positions" are always treated as bytes and not as characters.
Cause php source files should be parsed not assuming any string encode parsers must deal always  with bytes sequences. :sweat_smile: so I will try to change my lexer again to treat php source code always as bytes sequences without decoding it!  
Thanks for your time and sorry for bothering you with this PR.

PR #222, opened 2018-02-12
Don't emit "this test did not perform any assertions"

TysonAndre:
See https://phpunit.de/manual/current/en/risky-tests.html

I'm using php 7.2.2, and phpunit 6.5.6 gets installed.
I believe that the newer phpunit releases were stricter by default about
tests that don't perform any assertions


roblourens:
I guess this is from invariants tests that don't necessarily assert anything.

Issue #185, opened 2017-09-20
Output of `token_get_all()` depends on php build and config options for short tags


TysonAndre:
```bash
$ php --no-php-ini -d short_open_tag=0 -r 'echo json_encode(token_get_all("<? php"));'
[[321,"<? php",1]]$                     
$ php --no-php-ini -d short_open_tag=0 -r 'echo json_encode(token_get_all("<? php")) . "\n";'
[[321,"<? php",1]]
$ php --no-php-ini -d short_open_tag=1 -r 'echo json_encode(token_get_all("<? php")) . "\n";'
[[379,"<?",1],[382," ",1],[319,"php",1]]
$ php --no-php-ini -d short_open_tag=0 -r 'echo json_encode(token_get_all("<?php?>")) . "\n";'
[[321,"<?php?>",1]]
$ php --no-php-ini -d short_open_tag=1 -r 'echo json_encode(token_get_all("<?php?>")) . "\n";'
[[379,"<?",1],[319,"php",1],[381,"?>",1]]
```

This seems to be related to the options used to build PHP 
See https://secure.php.net/manual/en/ini.core.php#ini.short-open-tag and https://secure.php.net/manual/en/language.basic-syntax.phptags.php

>  PHP also allows for short open tag <? (which is discouraged since it is only available if enabled using the short_open_tag php.ini configuration file directive, or if PHP was configured with the --enable-short-tags option). 

This also causes two test failures in programStructure21.php.tree and programStructure13.php.tree, and causes that test to save a different `.tree` to disk (Which may then get accidentally be configured)

- See programStructure21.php and programStructure13.php. 
- Maybe detect this ini setting and conditionally skip those tests`ini_get('short_open_tag') === 1` in my PHP build from source, which used travis/compile.sh

---

Impact:

- Test failure is inconvenient for developers working on project
- The tolerant-php-parser won't work as well on codebases with short open tags if a PHP binary that isn't configured to support short open tags is used to analyze it.
  (I don't think this is a common use case. This issue affects `<?`, not `<?=`)

Possible Remediations:

- ini_set() can't be used
- May be able to split the output of token_get_all() into smaller tokens and process the smaller tokens, if a token begins with `<?` (Might do the wrong thing for non-PHP)
- Document and accept this bug?
- Refuse to start server unless `short_open_tag` is disabled via ini setting (checkable via ini_get())
  (Or optionally restart binary under the hood)

TysonAndre:
[...]

PR #227, opened 2018-02-12
Workaround for #185 when running tests locally

TysonAndre:
This is annoying for developers running unit tests
with short_open_tag enabled
(e.g. if you build php from source with support for that)

- By making this change, this PR ensures that those developers
  don't accidentally commit modified versions of
  programStructure13.php.tree, .diag, etc.

The short_open_tag option cannot be changed at runtime.

An alternative could be to add short_open_tag=0 to phpunit.xml
(But in the future, someone might want to explicitly test the
behavior with short open tags?)

### DIFF
--- a/tests/ParserGrammarTest.php
+++ b/tests/ParserGrammarTest.php
@@ -79,10 +79,14 @@ class ParserGrammarTest extends TestCase {
     public function treeProvider() {
         $testCases = glob(self::FILE_PATTERN . ".php");
         $skipped = json_decode(file_get_contents(__DIR__ . "/skipped.json"));
+        $hasShortOpenTag = (bool)ini_get('short_open_tag');
 
         $testProviderArray = array();
         foreach ($testCases as $testCase) {
             if (in_array(basename($testCase), $skipped)) {
+                continue;
+            }
+            if ($hasShortOpenTag && in_array(basename($testCase), ['programStructure13.php', 'programStructure21.php'])) {
                 continue;
             }
             $testProviderArray[basename($testCase)] = [$testCase, $testCase . ".tree", $testCase . ".diag"];


roblourens:
I don't want to hardcode test names like that. I think that adding it to phpunit.xml would be better. Why do you have short tags enabled?

TysonAndre:
Since you're fine with the phpunit.xml override, I'm using that. I verified that LexicalGrammarTest continues to work

I'm just using the default php settings; I don't have any use cases for short tags.

I use https://github.com/php/php-src/blob/PHP-7.2/travis/compile.sh and rebuild php from source 

http://php.net/manual/en/configure.about.php#configure.options.misc (would either have to add the entry short_open_tags=0 to php.ini or --disable-short-tags when building PHP)

TysonAndre:
Never mind, phpunit.xml does count as being set at runtime, so it won't work. The php option for short_open_tag was already a constant when `vendor/bin/phpunit` was executed.

roblourens:
Can this be closed?

TysonAndre:
> Should this be closed?

I changed this back to the ugly solution of using `            if ($hasShortOpenTag && in_array(basename($testCase), ['programStructure13.php', 'programStructure21.php'])) {`.

The commend I made on Feb 15 was just to mention why I couldn't have the clean solution you proposed of using phpunit.xml. The solution in this PR won't affect the majority of PHP installations (13 and 21 will continue to run), and will avoid confusion (i.e. accidentally checked in expectation files) when reviewing PRs made by people with uncommon PHP settings. Then again, based on the PRs to this repo so far, I might be the only contributor using php-src/travis/compile.sh, so closing this is probably fine.

Issue #19, opened 2017-01-18
LHS of assignment must be variable


nikic:
Noticed while looking at [example 4](https://github.com/Microsoft/tolerant-php-parser/blob/master/docs/HowItWorks.md#example-4):

In PHP an expression like `$a == $b = $c` is parsed as `$a == ($b = $c)`, because this is the only valid way to parse the code under the constraint that the LHS of an assignment must be a variable. The parser currently treats this as `($a == $b) = $c` instead.

nikic:
The same also applies to other expressions that accept only variables on the LHS, for example PHP interprets `!$x instanceof Y` as `!($x instanceof Y)`, while the parser currently treats it as `(!$x) instanceof Y`.

mousetraps:
Ahh, makes sense. Thanks for pointing that out! I was wondering why variables were called out separately from expressions in general.

For `instanceof`, I think I'm missing something because both the spec and yacc grammar seem to specify expression as LHS. 

Looking back at my notes, my interpretation of the spec was that `unary-expression` takes precedence over `instanceof-expression`, which in turn takes precedence over `multiplicative-expression`, where both intanceof and multiplicative expressions are defined similarly, which led me to believe they're parsed similarly.  

```
instanceof-expression:
   unary-expression
   instanceof-subject   instanceof   instanceof-type-designator

instanceof-subject:
   expression

multiplicative-expression:
   instanceof-expression
   multiplicative-expression   *   instanceof-expression
   multiplicative-expression   /   instanceof-expression
   multiplicative-expression   %   instanceof-expression
```

Should `instanceof-subject` be variable, rather than `expression`? And if so, how exactly does that get specified in the yacc grammar? because i think I'm missing something.

nikic:
You're right about `instanceof`, I got confused there. The LHS is a normal expression, it's the RHS that is a (restricted) variable.

However, the spec is incorrect about the precedence of `instanceof`: It should have higher precedence than `!` (but lower than `~`, so different unary operator are treated differently here).

Hywan:
:+1:

mattacosta:
I also happened to come across the following example that is related to this while looking at an RFC:

`!$x = f()` should be `!($x = f())` instead of `(!$x) = f()`

mattacosta:
@TysonAndre The affected operators with this restriction are:
- All assignment operators (it looks like you got all of these)
- Increment and decrement operators

[...]

PR #228, opened 2018-02-12
For #19: Fix an edge case parsing `=` and `instanceof`

TysonAndre:
This does not fix every single edge case.
This only fixes the edge cases for unary operators
on the left hand side of binary operators expecting a variable.

- `if (!$x = expr())` is something I've seen frequently, which is why I'm fixing this here. A complete fix would probably be more complex (e.g. deeply nested binary operators)

(That probably isn't even every single binary operator or unary operator)

See https://github.com/Microsoft/tolerant-php-parser/issues/19#issue-201714377 and https://github.com/Microsoft/tolerant-php-parser/issues/19#issuecomment-282898236


TysonAndre:
Is there anything that needs to be addressed for this or PR 232?

roblourens:
Thanks for the PRs!

PR #229, opened 2018-02-12
Optimized readability and removed useless else after return

alexmanno:
Optimized readability and removed useless else after return without change logic

### DIFF
--- a/src/DiagnosticsProvider.php
+++ b/src/DiagnosticsProvider.php
@@ -72,7 +72,9 @@ class DiagnosticsProvider {
                 $token->start,
                 $token->getEndPosition() - $token->start
             );
-        } elseif ($token instanceof MissingToken) {
+        }
+
+        if ($token instanceof MissingToken) {
             return new Diagnostic(
                 DiagnosticKind::Error,
                 "'" .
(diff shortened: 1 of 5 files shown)


src/DiagnosticsProvider.php
```diff
@@ -72,7 +72,9 @@ private static function checkDiagnosticForUnexpectedToken($token) {
                 $token->start,
                 $token->getEndPosition() - $token->start
             );
-        } elseif ($token instanceof MissingToken) {
+        }
+
+        if ($token instanceof MissingToken) {
```
mousetraps: I understand where you're coming from - this change makes the code more "scannable" because it makes it easier to move your eyes straight down top to bottom, but I think readability is actually worse.

This change increases cognitive load, because you can no longer make the assumption that the logic in the `if` block is mutually exclusive, and didn't impact the second block in some unintended way. Falling into `elseif`, `else` blocks makes the logical flow clear - it makes it obvious that we don't execute the logic in the other blocks.

roblourens: I agree with @mousetraps (hey Sara!)

PR #232, opened 2018-02-16
Fixes #189 : Fix parsing edge cases in yield statements

### LINKED ISSUE #189: Parser should emit `null` for a yield with no expression, but currently emits MissingToken
The parser is incorrect on the function `<?php function() { yield; }`, which is syntactically valid generator. Expressions such as `$value = yield;` are also affected.
- The behavior of `yield from` does not need to be changed.
- The syntax for `yield;` is documented in https://secure.php.net/manual/en/language.generators.syntax.php#control-structures.yield.null

Expected: `elementValue` should be `null` instead of `MissingToken` for the above snippets.

The above cases were not part of the tests added in https://github.com/Microsoft/tolerant-php-parser/commit/7c4a59de153546c019abfdf100e45c26f76fce5e for #3

```
$ php dump.php '<?php function() { yield; }'
Node\SourceFileNode
    statementList: Node\Statement\ExpressionStatement
        expression: Node\Expression\AnonymousFunctionCreationExpression
            functionKeyword: Token: FunctionKeyword(134): "function"
            openParen: Token: OpenParenToken(203): "("
            closeParen: Token: CloseParenToken(204): ")"
            compoundStatementOrSemicolon: Node\Statement\CompoundStatementNode
                openBrace: Token: OpenBraceToken(205): " {"
                statements: Node\Statement\ExpressionStatement
                    expression: Node\Expression\YieldExpression
                        yieldOrYieldFromKeyword: Token: YieldKeyword(166): " yield"
                        arrayElement: Node\ArrayElement
                            elementValue: MissingToken: Expression(329): "" <-- This should be null
                    semicolon: Token: SemicolonToken(237): ";"
                closeBrace: Token: CloseBraceToken(206): " }"
        semicolon: MissingToken: SemicolonToken(237): ""
    endOfFileToken: Token: EndOfFileToken(1): ""
```

TysonAndre:
Update tests diagnostics and expected generated ASTs

For backwards compatibility, continue to make `yield from expr`
have an ArrayElement as a child node.

To maintain the invariant that all Nodes must have at least one child,
make $yieldExpression->arrayElement be null when parsing `yield;`

Aside: `yield &$a;` is (and should be)
parsed as the binary operator `(yield) & ($a)`.
This is surprising, but makes sense, being the only sane parse tree.

Verified with the PHP module nikic/php-ast (generates AST_BINARY_OP)

```php
var_export(ast_dump(
    ast\parse_code('<?php function test() { yield & $x; }', 50)
));
```

Misc: Fix wrong param name in doc.


TysonAndre:
Made $yieldExpression->arrayElement be null for `yield;`, that was accidentally left out.

Added unit tests for `yield & &$x` (invalid) and `yield &&$x` (valid use of `&&` operator in php 7)

PR #233, opened 2018-02-16
Fixes #188: Fix parsing empty variables

### LINKED ISSUE #188: Unexpected outputted AST parsing `<?php $x = $;`
( `<?php $x = $;` would be a literal with the right hand side missing a variable name)

The parser returns a Node\Expression\Variable where the `name` property is also an instance of `Node\Expression\Variable`. This seems wrong, and makes it harder to check if something is an incomplete variable name.

Observed (using custom shorthand dumper):

```
Node\SourceFileNode
    statementList: Node\Statement\ExpressionStatement
        expression: Node\Expression\AssignmentExpression
            leftOperand: Node\Expression\Variable
                name: Token: VariableName(3): "$a" 
            operator: Token: EqualsToken(238): " =" 
            rightOperand: Node\Expression\Variable
                dollar: Token: DollarToken(217): " $" 
                name: Node\Expression\Variable
                    name: MissingToken: VariableName(3): ""  <- The right hand side is a Variable, but it also has a Variable as `name`, but MissingToken of kind `VariableName` seems more reasonable
        semicolon: Token: SemicolonToken(237): ";"
    endOfFileToken: Token: EndOfFileToken(1): ""
```

Expected:

```
Node\SourceFileNode
    statementList: Node\Statement\ExpressionStatement
        expression: Node\Expression\AssignmentExpression
            leftOperand: Node\Expression\Variable
                name: Token: VariableName(3): "$a" 
            operator: Token: EqualsToken(238): " =" 
            rightOperand: Node\Expression\Variable
                dollar: Token: DollarToken(217): " $" 
                name: MissingToken: VariableName(3): "" <- want to flatten this?
        semicolon: Token: SemicolonToken(237): ";"
    endOfFileToken: Token: EndOfFileToken(1): ""
```

TysonAndre:
Previously, the fallback case when something beginning
with the token `$` wasn't a simple variable
would be to assume it would be a complex variable
such as `$$x`

Check that the first token would be valid before recursing.


roblourens:
Thanks!

PR #234, opened 2018-02-16
Fix bugs and docs of FilePositionMap

TysonAndre:
I wasn't using some of those APIs for Node/Token,
and didn't notice the incorrect phpdoc/code.

### DIFF
--- a/src/FilePositionMap.php
+++ b/src/FilePositionMap.php
@@ -24,7 +24,7 @@ class FilePositionMap {
     /** @var int the 0-based byte offset of the most recent request for a line number. */
     private $currentOffset;
 
-    /** @var int the line number for $this->currentOffset (updated whenever currentOffset is updated) */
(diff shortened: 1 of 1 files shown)


src/FilePositionMap.php
```diff
@@ -86,19 +86,23 @@ public function getEndLine($node) : int {
      * Similar to getStartLine but includes the column
      */
     public function getEndLineCharacterPositionForOffset($node) : LineCharacterPosition {
-        return $this->getLineCharacterPositionForOffset($node);
+        return $this->getLineCharacterPositionForOffset($node->getEndPosition());
```
roblourens: The other `ForOffset` methods take an actual offset, should this one have a different name?

roblourens: Also can `$node` have a type annotation?

TysonAndre:
```php
     /**
      * @param Node|Token $node
      * Similar to getStartLine but includes the column
      */
```

Renamed this method, the original name was incorrect

PR #244, opened 2018-05-12
Fixes #103: Support parsing multiple exception types

### LINKED ISSUE #103: Support parsing multiple exception types in catch clauses

TysonAndre:
Add a new array `otherQualifiedNameList` alongside qualifiedName
to `CatchClause`. (Contains remaining `Token`s and `QualifiedName`s)
(This design breaks backwards compatibility as little as possible)

- A future backwards incompatible release should merge the two
  properties into `qualifiedNameList`,
  or something along those lines.

Add a new helper method to parse the catch list.

- The new helper is required because
  `tests/cases/php-langspec/exception_handling/odds_and_ends.php`
  should not fail because `catch (int $e)` is **syntactically** valid php
  (in 7.x), i.e. `php --syntax-check` doesn't reject it.


TysonAndre:
Could you publish a 0.0.12 release tag if this is merged?

This will help when using this library with PHP 7.1+ codebases.

TysonAndre:
Do the maintainers have any thoughts on this PR? E.g. documentation or additional tests that should be added? 

If this is merged, https://github.com/Microsoft/tolerant-php-parser/pull/107 can be closed.

 I think this should still become a list, just in a backwards incompatible change to 1.0 or 0.1 (e.g. the dependency https://github.com/felixfbecker/php-language-server/blob/master/composer.json#L28 of php-language-server would break if a name became a list of names)

> A future backwards incompatible release should merge the two
> properties into qualifiedNameList,
> or something along those lines.

PR #252, opened 2018-07-28
RFC: Static analysis fixes

dantleech:
This PR introduces PHPStan and fixes the code up to level 2 (out of 7).

- [x] Various fixes for static analysis
- [x] Introduces PHPStan in the Travis build

There remain 84 errors at `--level=7` (there were 117 before this PR)

--

PHPStan is `>=7.1` only, in `.travis.yml` it (should be) configured to run only on the 7.2 build.

### DIFF
--- a/src/Node.php
+++ b/src/Node.php
@@ -13,6 +13,8 @@ use Microsoft\PhpParser\Node\Statement\NamespaceDefinition;
 use Microsoft\PhpParser\Node\Statement\NamespaceUseDeclaration;
 
 abstract class Node implements \JsonSerializable {
+    const CHILD_NAMES = [];
+
     /** @var array[] Map from node class to array of child keys */
     private static $childNames = [];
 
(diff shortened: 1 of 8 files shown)


src/Node.php
```diff
@@ -46,23 +48,29 @@ public function getStart() : int {
...
                     }
                     $child = $child[0];
                 }
+
                 if ($child instanceof Node) {
                     return $child->getFullStart();
-                } elseif ($child instanceof Token) {
+                }
+
+                if ($child instanceof Token) {
```
dantleech: Not strictly necessary...

src/Node.php
```diff
@@ -46,23 +48,29 @@ public function getStart() : int {
...
                     return $child->fullStart;
                 }
+
                 throw new \Exception("Unknown type in AST: " . \gettype($child));
             }
         };
 
-        throw new \Exception("Unknown type in AST: " . \gettype($child));
+        throw new \RuntimeException("Could not resolve full start position");
```
dantleech: This exception assumed that `$child` was defined, and it seems misleading. The new message is broadly more accutate but lacks detail. In anycase I think this is a `Should never happen` exception.

PR #258, opened 2018-09-02
Fix a notice in Parser->parseTernaryExpression()

### LINKED ISSUE #257: Notice emitted when analyzing `<?php $x = ?;` in parseTernaryExpression
```
vendor/microsoft/tolerant-php-parser/src/Parser.php:2651 [8] Undefined property: Microsoft\PhpParser\MissingToken::$parent
```

Stack trace for a similar crash:

```
#0  custom_error_handler() called at [/path/to/vendor/microsoft/tolerant-php-parser/src/Parser.php:2651]
#1  Microsoft\PhpParser\Parser->parseTernaryExpression() called at [/path/to/vendor/microsoft/tolerant-php-parser/src/Parser.php:1766]
#2  Microsoft\PhpParser\Parser->parseBinaryExpressionOrHigher() called at [/path/to/vendor/microsoft/tolerant-php-parser/src/Parser.php:1771]
#3  Microsoft\PhpParser\Parser->parseBinaryExpressionOrHigher() called at [/path/to/vendor/microsoft/tolerant-php-parser/src/Parser.php:1565]
#4  Microsoft\PhpParser\Parser->Microsoft\PhpParser\{closure}() called at [/path/to/vendor/microsoft/tolerant-php-parser/src/Parser.php:1554]
```

(It crashes because it has a MissingToken instead of a Node because nothing was before the `?`)

Expected: This should not emit notices, and the parent node being set should be valid.
Observed: This emits a notice.
Impact: This may crash language servers if the notice is emitted to stdout (and the transport used is stdin/stdout), or if the language server error handler terminates the program.

TysonAndre:
Fixes #257

### DIFF
--- a/src/Node/Expression/TernaryExpression.php
+++ b/src/Node/Expression/TernaryExpression.php
@@ -11,7 +11,7 @@ use Microsoft\PhpParser\Token;
 
 class TernaryExpression extends Expression {
 
-    /** @var Expression */
+    /** @var Expression|Token (only a token when token before '?' is invalid/missing) */
     public $condition;
 
     /** @var Token */
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -1763,7 +1763,7 @@ class Parser {
             }
 
             $leftOperand = $token->kind === TokenKind::QuestionToken ?
-                $this->parseTernaryExpression($leftOperand, $token) :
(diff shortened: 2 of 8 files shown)


roblourens:
CI error looks like a network glitch...

PR #260, opened 2018-09-23
Fix a PHP notice generating a scoped property access expression

TysonAndre:
And add a unit test

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -1236,7 +1236,7 @@ class Parser {
                 $node->addElement($delimeterToken);
             }
             $token = $this->getCurrentToken();
-            // TODO ERROR CASE - no delimeter, but a param follows
+            // TODO ERROR CASE - no delimiter, but a param follows
         } while ($delimeterToken !== null);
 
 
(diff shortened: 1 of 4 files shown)


src/Parser.php
```diff
@@ -2594,12 +2594,18 @@ private function parseMemberAccessExpression($expression):MemberAccessExpression
...
+     * @param Node|null $fallbackParentNode (Workaround for the invalid AST `use TraitName::foo as ::x`)
+     */
+    private function parseScopedPropertyAccessExpression($expression, $fallbackParentNode): ScopedPropertyAccessExpression {
         $scopedPropertyAccessExpression = new ScopedPropertyAccessExpression();
-        $scopedPropertyAccessExpression->parent = $expression->parent;
-        $expression->parent = $scopedPropertyAccessExpression;
+        $scopedPropertyAccessExpression->parent = $expression->parent ?? $fallbackParentNode;
+        if ($expression instanceof Node) {
+            $expression->parent = $scopedPropertyAccessExpression;
+            $scopedPropertyAccessExpression->scopeResolutionQualifier = $expression; // TODO ensure always a Node
```
roblourens: Is the TODO now resolved, since it's in `instanceof Node`?

TysonAndre:
In the negation of the if statement, it's null (not a Node) - I wasn't completely sure what should be done there, so I left the TODO as is for now.

PR #264, opened 2018-10-31
Move PHPStan configuration to its file

carusogabriel:
With this patch, we can make the usage of PHPStan both locally and in our CI easier. Instead of figure out what paths and level we're using in the project, one can just do the following now:

```bash
$ composer require --dev phpstan/phpstan
$ vendor/bin/phpstan analyze
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_script:
 
 script:
   - composer validate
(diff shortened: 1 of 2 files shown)


carusogabriel:
@roblourens To be a benchmark for https://github.com/nikic/PHP-Parser :)

PR #268, opened 2018-12-05
Bug fixes for parsing associativity

TysonAndre:
This fixes the most common cases for #19
(there may be other edge cases I haven't considered yet)

This also fixes incorrect precedence for `<=>`.
It has the same precedence as `==` -
https://secure.php.net/manual/en/language.operators.precedence.php
has correct information for the `<=>` operator.


TysonAndre:
I ran a test of all valid combinations of a subset of binary operators (according to what php-ast said was valid) to check for false positives. (not including assignment by ref)

This change to Parser.php brought the number of incorrectly parsed ASTs from 311 down to 4. https://github.com/TysonAndre/phan/compare/tests-fallback-only?targetBranch=master

(The 4 remaining edge  cases include `$x ?? $y ?: $z` and `$x ?: $y ?: $z`, which were pre-existing. The other two were nonsense with `instanceof` that shouldn't occur in valid code.)

roblourens:
Sorry, I will look at it soon. Been busy/on vacation.

TysonAndre:
This is part of the standard that most programmers normally don't need to think about.

The PHP manual mentions this (replace `==` with `<=>` in that sentence)

> Operators of equal precedence that are non-associative cannot be used next to each other, for example 1 < 2 > 1 is illegal in PHP. The expression 1 <= 1 == 1 on the other hand is legal, because the == operator has lesser precedence than the <= operator.

The below two groups have the same precedences within the groups:

- (non-associative) `< <= > >=`
- (non-associative) `== != === !== <> <=>`

roblourens:
That's kind of crazy but makes sense, thanks!

PR #280, opened 2019-02-10
Use fully qualified substr in Token.php

TysonAndre:
Token->getText() will be called frequently in some applications.

This is slightly more efficient; the php interpreter can cache the address of the C function object when the namespace of the function is unambiguous.


roblourens:
Thanks!

PR #281, opened 2019-02-20
Support parsing `??=` (only supports php 7.4)

### LINKED ISSUE #277: PHP 7.4-dev coalesce assign support (??=)
Filing this to keep track, this can wait until alphas are out.

`'??='` is a new token, similar to `+=` 

https://wiki.php.net/rfc/null_coalesce_equal_operator

TysonAndre:
This is due to limitations of the lexer and token_get_all().
It's possible to manually add compatibility
by joining adjacent `??` and `=` tokens if `??=`
occurs anywhere in the file contents.

- I'm not sure if this project would choose to do that.
  Right now, it doesn't have a way to provide the target PHP version.
- T_COALESCE_EQUAL is a new constant; add a substitute value for older PHP
  versions

Fixes #277


roblourens:
So currently this would parse it properly when using PHP 7.4 and not parse it otherwise, right? That sounds fine with me for now.

TysonAndre:
> So currently this would parse it properly when using PHP 7.4 and not parse it otherwise, right? That sounds fine with me for now.

That's correct, token_get_all() returns `??` and `=` tokens in php <= 7.3, and tolerant-php-parser would continue to skip over those tokens in php <= 7.3

PR #282, opened 2019-02-23
Properly parse @$x += $y as @($x += $y)

TysonAndre:
Fixes one edge case in #19

Also, fix parsing of `@$x instanceof $y`.
My earlier PR fixed `!$x instanceof $y` but broke that.


TysonAndre:
ok, fixed

PR #287, opened 2019-03-10
Optimizations to PhpTokenizer

TysonAndre:
Speeds up parsing speed by a few percentage points.

- Move skipping prefix out of the loop.
  It isn't used except when getting doc comments, which is much less
  common than getting all tokens
- Handle strings separately from arrays in token_get_all.
- Move `if` within a switch to its own case.


roblourens:
I think the CI error is from this change, can you investigate? I think that entry in phpstan.neon can just be removed.

roblourens:
Thanks!

PR #293, opened 2019-05-12
Support parsing PHP 7.4's arrow functions

### LINKED ISSUE #291: Support PHP 7.4 arrow functions (fn($x) => expr)
These have similar properties to closures, except:

- There is a single expression instead of a function body
- There is no use clause, it's implicit

More details can be found in https://wiki.php.net/rfc/arrow_functions_v2

> Arrow functions have lowest precedence. This means that the expression to the right of => will be consumed as far as possible:

Examples:

```
fn(array $x) => $x;
fn(): int => $x;
fn($x = 42) => yield $x;
fn(&$x) => $x;
fn&($x) => $x;
fn($x, ...$rest) => $rest;
static fn() => 1;
```

-----

Ideas on how this would be represented:

Create a new ArrowFunctionCreationExpression with staticModifier, FunctionHeader, FunctionReturnType, arrowToken, and resultExpression: Token|Node (instead of FunctionBody). There would be no FunctionUseClause.

It would have these properties in this order (when the ast is valid):

- staticModifier: Token|null
- functionKeyword: Token (to reuse FunctionName, but `fn` instead of `function`)
- byRefToken: Token|null
- openParen: Token
- parameters:  DelimitedList\ParameterDeclarationList
- closeParen: Token
- colonToken: Token|null
- questionToken: Token|null
- returnType: Token|qualifiedName
- arrowToken: Token (`=>`)
- resultExpression: Token|Node

TysonAndre:
This requires PHP 7.4-dev from May 2nd or later.

Add various test cases of this working,
as well as tests of error tolerance.

Fixes #291


TysonAndre:
@roblourens  - Any feedback for this and #294 ?

TysonAndre:
@roblourens - Any feedback for this and #294 and #295 ?

PR #299, opened 2019-09-11
Test with php 7.4, update expected parse trees.

TysonAndre:
The tests in the parser74 folder are only run in php 7.4.
Some tokens such as `fn` are only generated when php 7.4's lexer is used.

My earlier PR added dotDotDotToken to arrayElement because php 7.4 supported the array spread operators in arrays, but I didn't test with 7.4.


TysonAndre:
@roblourens - thoughts?

PR #301, opened 2019-10-21
Group tokens logic 

gnoe:
Hi,
Can you check if this changes may help?
In the Go [go/token](https://github.com/golang/go/blob/master/src/go/token/token.go) package. Tokens are arranged by groups, "keywords", "literal", "operators" and this tokens are delimited by constants like "begin_keyword" ... "end_keyword", so any token between those constants are keywords. So,  to check if a token is keyword you don't have to iterate every token considered as a "keyword" and only check the method: token.isKeyword() => begin_keyword < tok && tok < end_keyword.

So, I decided to use this approach to organize the tokens of the parser and create helper functions to decide if a token is a keyword, the start of an expression, a parameter type, or anything that may help us in the parser.
using the script in tools/GenerateTokens.php which use the file tokens.php in the same folder, we can generate the TokenKind.php file if necesary. With this change methods like isStatementStart (9 lines) or isExpressionStart(10 lines) looks smaller in the parser and things like checking for in_array($token->kind, $this->keywordTokens) can be replaced with $token->isReserved() where isReserved only checks that the token is in the range of the keywords.
I created a TokensTest and I ran all tests before and after. It looks like that at least I didn't break anything.

I would appreciate any comment on this.
Thank you!

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -1319,7 +1319,7 @@ class Parser {
      *
      * @param int $kind
      */
-    private function lookAhead1($kind)
+    private function lookahead1($kind)
     {
         $startPos = $this->lexer->getCurrentPosition();
         $startToken = $this->token;
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -473,7 +473,7 @@ class Parser {
 
                 // labeled-statement
                 case TokenKind::Name:
-                    if ($this->lookahead(TokenKind::ColonToken)) {
+                    if ($this->lookahead1(TokenKind::ColonToken)) {
                         return $this->parseNamedLabelStatement($parentNode);
                     }
                     break;
@@ -524,7 +524,7 @@ class Parser {
                 // class-declaration
                 case TokenKind::FinalKeyword:
                 case TokenKind::AbstractKeyword:
-                    if (!$this->lookahead(TokenKind::ClassKeyword)) {
+                    if (!$this->lookahead1(TokenKind::ClassKeyword)) {
                         $this->advanceToken();
                         return new SkippedToken($token);
                     }
@@ -537,7 +537,7 @@ class Parser {
 
                 // namespace-definition
                 case TokenKind::NamespaceKeyword:
-                    if (!$this->lookahead(TokenKind::BackslashToken)) {
+                    if (!$this->lookahead1(TokenKind::BackslashToken)) {
                         // TODO add error handling for the case where a namespace definition does not occur in the outer-most scope
                         return $this->parseNamespaceDefinition($parentNode);
                     }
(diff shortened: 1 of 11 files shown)


gnoe:
The branch name is wrong, sorry.

src/TokenKind.php
```diff
@@ -10,205 +10,308 @@ class TokenKind {
...
+    }
+
+    public static function isKeyword(int $tokenKind): bool
+    {
+        return 6 < $tokenKind && $tokenKind < 201;
+    }
+
+    public static function isParameterTypeDeclaration(int $tokenKind): bool
+    {
+        return (101 < $tokenKind && $tokenKind < 104) || (203 < $tokenKind && $tokenKind < 209);
```
TysonAndre: The constants for token kinds vary between PHP versions. What works in PHP 7.0 likely won't work in PHP 8.0 and not in PHP 7.4. There's also no guarantee that php-src wouldn't reorder token kinds.

src/Parser.php
```diff
@@ -806,104 +756,23 @@ private function isExpressionStart($token) {
...
-                case TokenKind::UnsetCastToken:
-
-                // anonymous-function-creation-expression
-                case TokenKind::StaticKeyword:
-                case TokenKind::FunctionKeyword:
-                case TokenKind::FnKeyword:
-                    return true;
             }
-            return \in_array($token->kind, $this->reservedWordTokens, true);
+            return $token->isReserved();
```
TysonAndre: I assume end-users would care more about the performance of parsing than the design - function calls are fairly slow in php, and replacing in_array with a call would likely hurt performance.

Replacing with checking for set membership would be a tiny bit faster.

Also, is ArrayKeyword not reserved?

TysonAndre: A question this raises is whether it'd be useful to provide these helpers on Token to end-users.

What use cases did you have in mind for applications using tolerant-php-parser with isReserved()?

gnoe: I used the TokenStringMaps as reference, the same Tokens in KEYWORDS and RESERVED_WORDS are in the isKeyword() and isReserved(). 

Perhaps the Token class is not the right place to use it, at the end, in the Token class these methods are only shortcuts to TokenKind::is...(int $kind) and these may be relevant to the Parser only.

My initial goal was to rely on these methods instead of Parser's properties:
- reservedWordTokens
- keywordTokens
- nameOrKeywordOrReservedWordTokens
- nameOrReservedWordTokens
- nameOrStaticOrReservedWordTokens
- parameterTypeDeclarationTokens
- returnTypeDeclarationTokens

And shrink methods isExpressionStart and isStatementStart, which are called many times during parsing.

PR #303, opened 2019-11-10
Support php 8.0 union types for param/property/return

### LINKED ISSUE #302: Support PHP 8.0 Union types RFC
https://wiki.php.net/rfc/union_types_v2 was merged into PHP master today. The grammar is https://wiki.php.net/rfc/union_types_v2#type_grammar

Union types of 2 or more types can now occur on

- Properties
- Parameters
- Return types

For backwards compatibility of existing applications, this could be supported similarly to 59b37a8d259496e1a3ad76381d809c505150c127 - parse the first type in the union type (unless it started with `?`, optionally), then parse any remaining types into a different property

TysonAndre:
To avoid crashing existing applications such as language servers,
the existing typeDeclaration properties continue to be of type
Node|Token.

An additional otherTypeDeclarations property was added to support union
types. Normally, it will either be null, or contain alternating
BarTokens and nodes/tokens for types.

Add various tests of union types in parameters, properties, and return
types.

Fixes #302


TysonAndre:
@roblourens - This is ready for another look.

PR #314, opened 2020-02-15
Support parsing `new (call_function())()`

TysonAndre:
And add tests for arbitrary parenthesized expressions in `new` and
`instanceof`


src/Parser.php
```diff
@@ -2822,19 +2822,37 @@ private function parseScopedPropertyAccessExpression($expression, $fallbackParen
         return $scopedPropertyAccessExpression;
     }
 
+    public function isParsingObjectCreationExpression($expression) {
```
roblourens: Is it typical in PHP to have a method and property with the same names? It threw me off.

PR #317, opened 2020-03-04
Add a way to override the lexer in a Parser instance

TysonAndre:
Currently, it's possible to do,
but would require a lot of code duplication and pinning to an exact
version of tolerant-php-parser, which discourages overriding the lexer.

I can think of the following use cases for this.
My main reason is to reuse token_get_all() elsewhere,
but being able to parse `T_FN` in php < 7.4 (for short arrow functions) is also convenient.

- Multiple applications needing to use the result of token_get_all
  for the same file. If none of them modify the array,
  it's much faster to reuse the same array than to create this. (benchmarking, the time savings aren't as large as I thought, but there might be some memory savings for reusing the strings in the array)

  For example, Phan's language server mode will potentially use
  tolerant-php-parser. In addition to that, it also uses token_get_all
  in InlineHTMLPlugin (to check for misuse of inline HTML) and sometimes
  in BuiltinSuppressionPlugin
  (to list T_COMMENT/T_DOC_COMMENT containing `@phan-suppress-*`)

  Aside: https://wiki.php.net/rfc/token_as_object would be faster and
  more memory efficient than token_get_all() in php 8.0, if it gets approved
- Needing to call tolerant-php-parser on the same token stream,
  multiple times.
  (e.g. an application that modifies the Microsoft\PhpParser\Node
  instances (but not tokens), or which creates data structures from the
  original Node but usually discards them to save memory)

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -136,6 +136,21 @@ class Parser {
     }
 
     /**
+     * This method exists so that it can be overridden in subclasses.
+     * Any subclass must return a token stream that is equivalent to the contents in $fileContents for this to work properly.
+     *
+     * Possible reasons for applications to override the lexer:
+     *
+     * - Imitate token stream of a newer/older PHP version (e.g. T_FN is only available in php 7.4)
+     * - Reuse the result of token_get_all to create a Node again.
+     * - Reuse the result of token_get_all in a different library.
+     */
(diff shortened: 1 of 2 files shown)


TysonAndre:
This would also be useful for `T_MATCH` in https://github.com/microsoft/tolerant-php-parser/pull/326 . @roblourens - thoughts?

PR #318, opened 2020-04-11
Support using `throw` as an expression

TysonAndre:
To maximize backward compatibility, continue using ThrowStatement for
throw being the first keyword in a statement.

For https://wiki.php.net/rfc/throw_expression
which is currently in the voting phase, but is likely to pass
(should only get merged after the implementation is merged into php 8.0)


TysonAndre:
This is ready to be looked at -`throw` expression support has been merged into 8.0-dev.

See `https://github.com/php/php-src/pull/5279#issuecomment-618450642`

TysonAndre:
@roblourens - thoughts?

roblourens:
Thanks!

PR #324, opened 2020-06-14
Support php 8.0 constructor property promotion

### LINKED ISSUE #323: Support php 8.0 constructor property promotion
https://wiki.php.net/rfc/constructor_promotion

For example, add an optional visibility token to parameters so that this can be parsed.

Applications using tolerant-php-parser can check for those and indicate if the usage is an error (i.e. not in __construct of a class)

TysonAndre:
https://wiki.php.net/rfc/constructor_promotion

Note that PHP 8.0 will parse visibility modifiers on all parameters,
but it will be a compile-time error to use them anywhere except
a __construct method.

Fixes #323


tests/cases/parser/parameterPromotion2.php.diag
```diff
@@ -0,0 +1,50 @@
+[
```
roblourens, line 1: Is `parameterPromotion2.php` missing?

TysonAndre:
> Thanks! Is this good to be merged even though PHP 8 is not released? Is that proposal stable?

I'd consider it to be stable. The php-src PR linked in https://wiki.php.net/rfc/constructor_promotion#php_rfcconstructor_property_promotion was already merged to the 8.0-dev(master) branch, and there's no discussion of changing it on the mailing list. The release managers plan to release 8.0.0 in November 2020 (https://wiki.php.net/todo/php80#timetable)

PR #326, opened 2020-07-04
Support match expression v2 RFC

TysonAndre:
https://wiki.php.net/rfc/match_expression_v2 was approved.

Note that the default keyword and expressions can't be mixed in a
condition list, but this is an error-tolerant parser,
so that check wasn't added.

This was tested against 8.0-dev


TysonAndre:
This functionality has been merged into php 8.0-dev. The tests of `match` continue to work.

PR #328, opened 2020-07-26
Support testing with phpunit 7 as well

TysonAndre:
Prepare for testing with PHP 8.0. This works well enough with `--ignore-platform-reqs`

An alternative would be to drop support for php 7.0

The call to parent::addFailure was a no-op.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
         "php": ">=7.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.4"
+        "phpunit/phpunit": "^6.4|^7.5.20"
     },
     "license": "MIT",
     "authors": [
--- a/tests/CallbackTestListener.php
+++ b/tests/CallbackTestListener.php
@@ -1,0 +1,36 @@
+<?php
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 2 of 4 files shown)


TysonAndre:
@roblourens - the build failure was fixed and this is ready for a look. Alternately, dropping support for 7.0 and possibly 7.1 is possible as mentioned in #329, especially since 8.0.0 was released this week and it's hard to find libraries officially supporting both 7.0 and 8.0

TysonAndre:
```
» vendor/bin/phpunit tests/ParserGrammarTest.php

Warning: Private methods cannot be final as they are never overridden by other classes in /path/to/vendor/phpunit/phpunit/src/Util/Configuration.php on line 176

Fatal error: Cannot acquire reference to $GLOBALS in /path/to/tolerant-php-parser/vendor/phpunit/phpunit/src/Util/Configuration.php on line 543
» php --version
PHP 8.1.0-dev (cli) (built: Mar 27 2021 12:32:25) ( NTS DEBUG )
Copyright (c) The PHP Group
Zend Engine v4.1.0-dev, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.0-dev, Copyright (c), by Zend Technologies
```

Even with this, running tests in php 8.1 will require an even newer phpunit version - that might be done in tolerant-php-parser 1.0.0

roblourens (approved):
Thank you!

PR #331, opened 2020-08-02
Support php 8.0's "Treat namespaced names as single token" and "?->"

### LINKED ISSUE #327: Support PHP 8.0's "Treat namespaced names as single token" change
https://wiki.php.net/rfc/namespaced_names_as_token passed on July 31st. It makes the following changes to tokenization:

```
Foo\Bar;
// Before: T_STRING T_NS_SEPARATOR T_STRING
// After:  T_NAME_QUALIFIED
// Rule:   {LABEL}("\\"{LABEL})+
 
\Foo;
// Before: T_NS_SEPARATOR T_STRING
// After:  T_NAME_FULLY_QUALIFIED
// Rule:   ("\\"{LABEL})+
 
namespace\Foo;
// Before: T_NAMESPACE T_NS_SEPARATOR T_STRING
// After:  T_NAME_RELATIVE
// Rule:   "namespace"("\\"{LABEL})+
```

**I think the simplest way to support existing applications might be to manually split up the `T_NAME_*` tokens into T_STRING, T_NAMESPACE, T_NS_SEPARATOR, etc** after the call to token_get_all()

- Other new syntax like `T_MATCH` or `T_FN` can have the same thing done to emulate that. Switches are fast in newer php versions when all arguments are definitely integers.

### LINKED ISSUE #330: Support php 8.0's nullsafe operator (?->)
https://wiki.php.net/rfc/nullsafe_operator

```php
   $foo = $a?->b();
// --------------- chain 1
//        -------- chain 2
// If $a is null chain 2 is aborted, method b() isn't called, null is assigned to $foo
 
   $a?->b($c->d());
// --------------- chain 1
//        -------  chain 2
// If $a is null chain 1 is aborted, method b() isn't called, the expression `$c->d()` is not evaluated
 
   $a->b($c?->d());
// --------------- chain 1
//       --------  chain 2
// If $c is null chain 2 is aborted, method d() isn't called, null is passed to `$a->b()`
```

This appears to be straightforward - T_NULLSAFE_OBJECT_OPERATOR looks to have the same precedence and places it can occur as T_OBJECT_OPERATOR(`->`)

TysonAndre:
Without this, tolerant-php-parser won't work in php 8.0.

Fixes #327
Fixes #330

I confirmed that SWITCH_LONG opcodes (jumptable optimization) continued to be generated for getTokensArrayFromContent in this patch in php 7.4 and php 8.0 for `php -d opcache.opt_debug_level=0x20000 vendor/bin/phpunit 2>&1|less`

-----

Support php 8.0's nullsafe operator

See https://wiki.php.net/rfc/nullsafe_operator and the corresponding
github implementation PR.


TysonAndre:
@roblourens - thoughts?

TysonAndre:
@roblourens - pinging again because this PR fixes issues caused by a major change in lexing and the first php 8.0 beta is out - https://www.php.net/archive/2020.php#2020-08-06-3

src/PhpTokenizer.php
```diff
@@ -130,6 +131,57 @@ public static function getTokensArrayFromContent(
...
                     break;
+                case \PHP_VERSION_ID >= 80000 ? \T_NAME_QUALIFIED : -1000:
+                case \PHP_VERSION_ID >= 80000 ? \T_NAME_FULLY_QUALIFIED : -1001:
+                    // NOTE: This switch is called on every token of every file being parsed, so this traded performance for readability.
+                    //
+                    // PHP's Opcache is able to optimize switches that are exclusively known longs,
+                    // but not switches that mix strings and longs or have unknown longs.
+                    // Longs are only known if they're declared within the same *class* or an internal constant (tokenizer).
+                    //
+                    // For some reason, the SWITCH_LONG opcode was not generated when the expression was part of a class constant.
```
roblourens, line 142: Thanks for the explanation!

PR #333, opened 2020-08-07
Issue 178 - Declare Statement Directive List

### LINKED ISSUE #178: DeclareStatement hardcodes assumption there is a single value (low priority)
Low priority because I've never seen code like this in practice. 

```
<?php declare(strict_types=1, ticks=1);
echo "This can be parsed\n";
```

This causes the parser to emit various MissingToken in the output. Running the above script with `php` reports no issues.

PythooonUser:
## Summary
This PR adds support for declare directive lists. Closes #178.

Adds a valid parse tree for:
```php
declare(strict_types=1, ticks=1);
```

## Remaining Issues
1. I couldn't find any official specs that would indicate that the above is valid (despite PHP not returning any errors)
1. Other parser (e.g. intelephense) report erros here, i.e. for the encountered `,`: `Unexpected ','. Expected ')'.`
1. Test case `tests/cases/parser/declareStatement12.php.tree` breaks. That's expected. Using `DelimitedList` allows for an empty list. The behaviour of this test case is that it permits an empty `declare` block. Ideas? Related to #4? My proposal would be to update `DelimitedList` with an option to allow/disallow empty lists. But that might break existing logic?
1. Only three types of declare directives are currently allowed. Should that be highlighted with an error when the user tries to insert something else?
1. Only literals are considered valid values for the directives. Should that be highlighted with an error when the user tries to insert something else?
1. The new test case `tests/cases/parser/declareStatement14.php.tree` seems a but wordy. I could shrink it down by excluding the `echo` statement.


TysonAndre:
Hi - tolerant-php-parser is used by a lot of projects, and previous PRs have attempted to ensure that those versions would work even after an upgrade from 0.0.x to 0.0.(x+1). - https://github.com/microsoft/tolerant-php-parser/network/dependents?package_id=UGFja2FnZS01NDI3MTMyOTI%3D

For example, if those projects expected $declareStatement->declareDirective to exist (e.g. to check for strict_types), they would encounter notices or misbehave on files that previously worked.

In previous commits, e.g. to support union types, catch class lists, etc, the original datatype of the first list element was preserved, and a new property such as $otherDeclareDirectives was added to the properties (and class constant) that was the list of the remaining values. (and a TODO was added to unify them in the next major release)

```
» ag '\$other'
src/Node/FunctionReturnType.php
19:    public $otherReturnTypes;

src/Node/Parameter.php
23:    public $otherTypeDeclarations;

src/Node/MissingMemberDeclaration.php
24:    public $otherTypeDeclarations;

src/Node/CatchClause.php
26:    public $otherQualifiedNameList;

src/Node/PropertyDeclaration.php
28:    public $otherTypeDeclarations;
```

PythooonUser:
Thanks for the clarification! Makes total sense. I will have a look at these examples.

PythooonUser:
Alright, I made a quick fix to address the backwards compatibility of the changes and updated the existing tests accordingly. Let me know what you think :)

Issue #320, opened 2020-04-26
Support attributes v2 RFC when syntax is finalized


TysonAndre:
https://externals.io/message/110217  [RFC] Amendments to Attributes is proposing changing the final syntax

https://wiki.php.net/rfc/attributes_v2 is in the voting phase - there are currently 48 yes votes and 1 no vote, and `<<Attribute(optional arg list)>>` is the preferred syntax.

> This RFC proposes Attributes as a form of structured, syntactic metadata to declarations of classes, properties, functions, methods, parameters and constants.

Relevant parts of the CFG in the implementation PR in Zend/zend_language_parser.y

```c
attribute_arguments:
		expr
			{ $$ = zend_ast_create_list(1, ZEND_AST_ARG_LIST, $1); }
	|	attribute_arguments ',' expr
			{ $$ = zend_ast_list_add($1, $3); }
;

attribute_decl:
		class_name_reference
			{ $$ = zend_ast_create(ZEND_AST_ATTRIBUTE, $1, NULL); }
	|	class_name_reference '(' ')'
			{ $$ = zend_ast_create(ZEND_AST_ATTRIBUTE, $1, NULL); }
	|	class_name_reference '(' attribute_arguments ')'
			{ $$ = zend_ast_create(ZEND_AST_ATTRIBUTE, $1, $3); }
;

attribute:
		T_SL attribute_decl T_SR	{ $$ = $2; }
;

attributes:
		attribute				{ $$ = zend_ast_create_list(1, ZEND_AST_ATTRIBUTE_LIST, $1); }
	|	attributes attribute	{ $$ = zend_ast_list_add($1, $2); }
;
```

liyandh:
abcde/3

liyandh:
x^3+4x=7

PR #342, opened 2020-09-11
Also handle attributes in `new #[MyAttr] class {...}`, traits, interfaces

TysonAndre:
I forgot that anonymous classes were parsed into an
ObjectCreationExpression instead.

Also support attributes there, cause a diagnostic to be emitted if the
attribute group(s) aren't followed by the `class` keyword.

Add attributes support for TraitDeclaration and InterfaceDeclaration
in the same way it was added earlier for ClassDeclaration.

For #320


TysonAndre:
@roblourens - this is ready for review. I didn't notice anything else missing when working on an application using tolerant-php-parser

roblourens (approved):

PR #345, opened 2020-09-13
Fix handling of named label for goto.

### LINKED ISSUE #154: goto parsing fails with  empty target
Symfony's route generator generates a url matcher on disk like so:

```php
<?php
class UrlMatcher {
    public function match($path)
    {
        if ('/path' === $pathinfo) {
            if ('POST' !== $canonicalMethod) {
                $allow[] = 'POST';
                goto not_this_path;
            }
            return array ('_controller' => 'method',  '_route' => 'this_path');
        }
        not_this_path:
        // other routes with their own goto not_that_path
    }
}
```
PHP seems to allow empty targets, since it runs the code just fine.  This is a regression compared to the previous version of the language server.

TysonAndre:
The label is a standalone statement just like any other statement.
It isn't associated with the next statement by PHP's parser.

E.g. `while (false) label: echo "test\n";` echoes "test".

Fixes #154


TysonAndre:
Actually, I think it's incorrect for a goto label to be associated with a statement.

```
php > while(false) label: echo "test";
test
```

That example gets parsed equivalently to `while(false) { label: } echo "test";` by php, but tolerant-php-parser incorrectly treats it as `while (false) {label: echo "test"}`

Looking at the php-src parser implementation, a goto label is a standalone statement just like any other statement

```c
	|	';'	/* empty statement */ { $$ = NULL; }
	|	T_TRY '{' inner_statement_list '}' catch_list finally_statement
			{ $$ = zend_ast_create(ZEND_AST_TRY, $3, $5, $6); }
	|	T_GOTO T_STRING ';' { $$ = zend_ast_create(ZEND_AST_GOTO, $2); }
	|	T_STRING ':' { $$ = zend_ast_create(ZEND_AST_LABEL, $1); }
;
```

roblourens (approved):
Good catch! I wouldn't have expected that behavior.

Issue #191, opened 2017-09-22
Support the php 7.1 iterable type hint as a unique token kind?


TysonAndre:
https://secure.php.net/manual/en/migration71.new-features.php#migration71.new-features.iterable-pseudo-type

Didn't see "iterable" mentioned anywhere in the codebase or the issues.

E.g. when parsing this snippet,

```php
<?php
function process_iter(iterable $y) {}
```

 the below is observed:

Observed: Token->kind is Name for "iterable"
Desired: Token->kind is IterableKeyword(new type) for "iterable"

The proposed behavior is similar to the current behavior for the "callable" keyword.

```
Node\SourceFileNode
    statementList: Node\Statement\InlineHtml
        scriptSectionStartTag: Token: ScriptSectionStartTag(323): "<?php\n"
    statementList: Node\Statement\FunctionDeclaration
        functionKeyword: Token: FunctionKeyword(134): "function"
        name: Token: Name(2): " process_iter"
        openParen: Token: OpenParenToken(203): "("
        parameters: Node\DelimitedList\ParameterDeclarationList
            children: Node\Parameter
                typeDeclaration: Node\QualifiedName
                    nameParts: Token: Name(2): "iterable"   <-- want IterableKeyword to exist
                variableName: Token: VariableName(3): " $y"
        closeParen: Token: CloseParenToken(204): ")"
        compoundStatementOrSemicolon: Node\Statement\CompoundStatementNode
            openBrace: Token: OpenBraceToken(205): " {"
            closeBrace: Token: CloseBraceToken(206): "}"
    endOfFileToken: Token: EndOfFileToken(1): "\n"
```

PR #348, opened 2020-11-07
[Feature] Support the php 7.1 iterable type hint as a unique token kind

camilledejoye:
Hi,

This PR implements the feature suggested in #191 

There already was a tests using `iterable`, for now I reused it.
If you want a dedicated test that's not a problem, I will simply need a few information:
* Should I remove the `iterable` type from the current `arrowFunctionUnionTypes.php` test so that it will not be impacted by the changes ?
* Where should I put the new test ? `tests/cases/parser71/iterableType.php` ?
* Is the example provided in the issue enough or do you have something else in mind ?

### DIFF
--- a/src/TokenKind.php
+++ b/src/TokenKind.php
@@ -87,6 +87,7 @@ class TokenKind {
     const YieldFromKeyword = 167;
     const FnKeyword = 168;
     const MatchKeyword = 169;
+    const IterableKeyword = 170;
 
     const OpenBracketToken = 201;
     const CloseBracketToken = 202;
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -141,7 +141,8 @@ class Parser {
         $this->parameterTypeDeclarationTokens =
             [TokenKind::ArrayKeyword, TokenKind::CallableKeyword, TokenKind::BoolReservedWord,
             TokenKind::FloatReservedWord, TokenKind::IntReservedWord, TokenKind::StringReservedWord,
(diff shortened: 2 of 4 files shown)


roblourens (approved):
This looks good to me but would you take a look too @TysonAndre?

TysonAndre:
No objections from me.

Phan continues to work with this change. I haven't tested other applications such as php-language-server, but it doesn't have much special handling for types since type inference isn't a part of php-language-server.


```
// only references to ReservedWord in php-language-server/src/DefinitionResolver.php
        // TRUE / FALSE / NULL
        // Resolve true and false reserved words to Types\Boolean
        if ($expr instanceof Node\ReservedWord) {
            $token = $expr->children->kind;
            if ($token === PhpParser\TokenKind::TrueReservedWord || $token === PhpParser\TokenKind::FalseReservedWord) {
                return new Types\Boolean;
            }

            if ($token === PhpParser\TokenKind::NullReservedWord) {
                return new Types\Null_;
            }
        }
```

src/TokenKind.php
```diff
@@ -87,6 +87,7 @@ class TokenKind {
     const YieldFromKeyword = 167;
     const FnKeyword = 168;
     const MatchKeyword = 169;
+    const IterableKeyword = 170;
```
dantleech, line 90: I think this should rather be a `ReservedWord` than a `Keyword`?

Issue #329, opened 2020-07-26
Publishing a new major version (dropping support for older php versions, finishing TODOs)


TysonAndre:
1. This allows using the `void` and `object` and nullable types from php 7.2. https://www.php.net/supported-versions.php still officially supports 7.2 for security patches (some third parties backport patches on a best effort basis, e.g. when packaged for Linux)
2. This allows upgrading phpunit to support php 8.0. Currently, no versions of phpunit support php 8.0, and https://phpunit.de/supported-versions.html indicates phpunit 6 and 7 are no longer maintained. Future phpunit versions require adding `:void`, which means supporting multiple phpunit versions is inconvenient.
3. This allows making the backwards incompatible changes that were added in TODOs - e.g. converting exception classes to a list, making types a list for union type support, etc

Related to https://github.com/microsoft/tolerant-php-parser/pull/328 - it's possible to work around and run php 8.0 in Travis so keeping php 7.0 support is possible

I'm only familiar with a few of the packages depending on this https://github.com/microsoft/tolerant-php-parser/network/dependents?dependent_type=PACKAGE&package_id=UGFja2FnZS01NDI3MTMyOTI%3D

- php-language-server hasn't received commits recently, but the language server is widely used and the maintainer is active and responsive elsewhere. (requires php ^7.0)
  It may be useful to support new token types in 8.0, but staying on 0.0.x indefinitely might add tech debt.
- I use this for http://github.com/phan/phan/ as a polyfill/fallback parser (php ^7.2)
- There are some other major dependents like phptree that I'm less familiar with

roblourens:
Thanks for opening the issue on php-language-server. I am ok with removing support for 7.0 and 7.1 but would look for feedback since Felix might have some idea of how many PHP devs are not updating regularly.

TysonAndre:
There's still no feedback on https://github.com/felixfbecker/php-language-server/issues/790 after 5 months, and with the release of php 8.0.0, there's a better idea of what applications would use for a new major release of tolerant-php-parser

TysonAndre:
> Thanks for opening the issue on php-language-server. I am ok with removing support for 7.0 and 7.1 but would look for feedback since Felix might have some idea of how many PHP devs are not updating regularly.

@roblourens - some questions before starting work on this

1. Did you think there should be a new major release for tolerant-php-parser?
[...]

PR #349, opened 2020-12-15
Proposed changes for 0.1.0

TysonAndre:
Drop support for byRefToken in ArgumentExpression - it's a syntax error
in php 7.0+ and is prohibited in php 5.4+.

Convert the foo and otherFooList to fooList for union types in catch
name lists, parameters, properties, and return union types.

Remove ThrowStatement in favor of ThrowExpression for php 8.0.
Remove EchoExpression in favor of EchoStatement - echo cannot be used as
an expression.

Remove statement from NamedLabelStatement - it's always null since a
named label is a single statement by itself.

TODO: Consistently support MissingToken in a single-element list in cases where
a non-empty list is required, such as after `?` in types
(php 8.0 union types forbid combining `?` with multiple types)

TODO: Check for any regressions in diagnostics

TODO: Validate that this can be used with real projects if they are refactored to use the new API

Rename to getFullStartPosition and getStartPosition

Consistently end public API methods with Position.
Avoid the need for callers to check if an object is a Node/Token before
calling a method.

For https://github.com/microsoft/tolerant-php-parser/issues/329
**NOTE: It may be useful to create a new `v1` branch or other name for the newly created major version - I don't think it was created yet**


roblourens:
This all looks fine so far. Moved to a v0.1.0 branch. Will call it that for now, that still implies a backwards-incompatible change under semver for 0.x.y.

TysonAndre:
This is ready for review. Note that the target `v0.1.0` branch should be reset to the commit of the `main` branch to filter out commits that were already approved and merged (such as enum support)

Note that in `.travis.yml`, `8.0` becomes the number `8`, which gets treated like the string `'8'`, which typically means the latest stable 8.x release.

TysonAndre:
@roblourens thoughts?

roblourens:
Sorry @TysonAndre, I will look at it this week

src/MissingToken.php
```diff
@@ -6,11 +6,14 @@
...
 
+use ReturnTypeWillChange;
+
 class MissingToken extends Token {
     public function __construct(int $kind, int $fullStart) {
         parent::__construct($kind, $fullStart, $fullStart, 0);
     }
 
+    #[ReturnTypeWillChange]
     public function jsonSerialize() {
```
TysonAndre, line 17: https://wiki.php.net/rfc/internal_method_return_types adds support for tentative return types, and `:mixed` was later added to JsonSerializable's method in 8.1 in https://github.com/php/php-src/pull/7051 . This means that implementations will have to return something (or throw), and also that the absence of a return type will emit a warning.

- fixes `vendor/microsoft/tolerant-php-parser/src/Token.php:113 [8192] Declaration of Microsoft\PhpParser\Token::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed` (a warning, not an error)

Because php 7 is still supported, add an annotation to suppress the warning instead.

Eventually this may have to drop support for 7.x because the mixed type requires 8.0, but that's probably enough time, but that depends on how many 8.x minor releases there are, which is unknown

roblourens (approved):
Thanks for your work @TysonAndre, sorry for the slow review. I am happy with this. Is this ready to merge and publish as 0.1.0?

TysonAndre:
> Is this ready to merge and publish as 0.1.0?

Yes, it is

PR #350, opened 2020-12-23
Add support for PHP 8.0 mixed type

dantleech:


src/Parser.php
```diff
@@ -142,7 +142,7 @@ public function __construct() {
             [TokenKind::ArrayKeyword, TokenKind::CallableKeyword, TokenKind::BoolReservedWord,
             TokenKind::FloatReservedWord, TokenKind::IntReservedWord, TokenKind::StringReservedWord,
             TokenKind::ObjectReservedWord, TokenKind::NullReservedWord, TokenKind::FalseReservedWord,
-            TokenKind::IterableKeyword]; // TODO update spec
+            TokenKind::IterableKeyword, TokenKind::MixedReservedWord]; // TODO update spec
```
dantleech, line 145: Unrelated to this PR but I think that `IterableKeyword` should be `Reserved` or?

TysonAndre: Yeah, it should have been IterableReservedword like ObjectReservedWord - it isn't a separate keyword like T_ARRAY.

Since they've already created a release, you could add an `/** @deprecated */ const IterableKeyword = self::IterableReservedWord` in TokenKind

dantleech: ok, updated

src/TokenKind.php
```diff
@@ -87,7 +87,7 @@ class TokenKind {
     const YieldFromKeyword = 167;
     const FnKeyword = 168;
     const MatchKeyword = 169;
-    const IterableKeyword = 170;
+    /** @deprecated use IterableReservedWord */const IterableKeyword = 170;
```
TysonAndre, line 90:
```suggestion
    /** @deprecated use IterableReservedWord */
    const IterableKeyword = self::IterableReservedWord;
```

My earlier suggestion was shorthand - style guidelines I'm aware of for php have doc comments on separate lines.
Single-line comments are used for properties when the comment is just one line.

src/TokenKind.php
```diff
@@ -170,6 +170,8 @@ class TokenKind {
     const StringReservedWord = 320;
     const BoolReservedWord = 321;
     const NullReservedWord = 322;
+    const MixedReservedWord = 420;
```
TysonAndre, line 173:
```suggestion
    const MixedReservedWord = 340;
```

It seems like `// const DollarOpenCurly = 339;` was used in the past, and all of the other reserved words except for iterable were using 330, and 400 is for special kinds of tokens?

The values don't affect the behavior of the parser, and I didn't notice this when contributing or reviewing (iterable) in the past (e.g. https://github.com/microsoft/tolerant-php-parser/pull/245), but consistently using 3xx for reserved words as much as possible would be preferable

TysonAndre:
LGTM but I'm not a maintainer, it's possible the maintainer is on vacation

dantleech:
ok thanks, no rush :)

roblourens (approved):
Thanks!

PR #352, opened 2021-03-27
Support parsing PHP 8.1 enums

TysonAndre:
Supports parsing https://wiki.php.net/rfc/enumerations

These tests pass locally after modifying the local installation of
phpunit 6 to workaround php 8.1 incompatibilities for `private function` and
$GLOBALS

Also, fix placeholder for T_NULLSAFE_OBJECT_OPERATOR prior to php 8.0.
This didn't matter in practice because
token_get_all never returns the constant that would be used in a key
prior to php 8.0

Enums are similar to classes, but:
- They forbid instance properties (and static properties)
- They allow `case` statements, which are the only allowed instances of
  the object. Those may or may not have an associated int/string type.
  The parse grammar seems to allow parsing any type
  but the compiler enforces that it's empty, an integer, or a string.


roblourens (approved):
Thank you!

PR #355, opened 2021-06-22
Error when calling getNamespaceDefinition on SourceFileNode

### LINKED ISSUE #255: Error when calling getNamespaceDefinition on SourceFileNode
When calling `getNamespaceDefinition` on a `SourceFileNode` an exception is thrown:

```
Invalid tree - SourceFileNode must always exist at root of tree.
```

The function tries to [fallback](https://github.com/Microsoft/tolerant-php-parser/blob/master/src/Node.php#L600) to the `SouceFileNode` from any child node type, but fails when the node is itself a `SourceFileNode`.

PythooonUser:
Hopefully fixes #255.

Adds an additional initial check for the source file node, after which we start searching in the ancestors.

### DIFF
--- a/src/Node.php
+++ b/src/Node.php
@@ -584,7 +584,7 @@ abstract class Node implements \JsonSerializable {
      * @return NamespaceDefinition|null
      */
     public function getNamespaceDefinition() {
-        $namespaceDefinition = $this instanceof NamespaceDefinition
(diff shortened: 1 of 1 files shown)


roblourens (approved):

roblourens:
Thanks!

PR #357, opened 2021-07-09
Support php 8.1 readonly property modifier RFC

TysonAndre:
https://wiki.php.net/rfc/readonly_properties_v2 is likely to pass


src/Parser.php
```diff
@@ -3522,6 +3594,7 @@ private function isEnumMemberDeclarationStart($token) {
             case TokenKind::ProtectedKeyword:
             case TokenKind::PrivateKeyword:
             // case TokenKind::VarKeyword:
+            // case TokenKind::ReadonlyKeyword:
```
roblourens, line 3597: Remove?

tests/cases/parser/parameterPromotion2.php.diag
```diff
@@ -1,50 +1 @@
...
-    },
-    {
-        "kind": 0,
-        "message": "Unexpected '}'",
-        "start": 157,
-        "length": 1
-    }
-]
\ No newline at end of file
+[]
```
roblourens, line 1: Was all of this supposed to change?

TysonAndre: Yes, this is the new expected result. Previously, a lot of syntax errors resulted from the additional modifier when only a single modifier was allowed

```php
<?php
class X {
    // Starting in php 8.1, it is a compile error rather than a syntax error to have more than one visibility modifier
    public function __construct(
        public protected $var
    ) {}
}
```

````
» php --version
PHP 8.1.0-dev (cli) (built: Jul 11 2021 12:15:15) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.0-dev, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.0-dev, Copyright (c), by Zend Technologies
» php tests/cases/parser/parameterPromotion2.php 
Fatal error: Multiple access type modifiers are not allowed in /mnt/DATA/ifwe_programming/tolerant-php-parser/tests/cases/parser/parameterPromotion2.php on line 5
```

PR #363, opened 2021-11-27
Fix PHP 8.2 deprecation notices in tolerant-php-parser

TysonAndre:
https://wiki.php.net/rfc/deprecate_dynamic_properties
was approved and merged into php 8.2.

Dynamic (undeclared) properties will be forbidden without
`#[AllowDynamicProperties]` in the next major release (9.0).

A separate followup question would be whether to include
`#[AllowDynamicProperties]` on the Node class declaration
or to instead do something similar to nikic/php-parser and add a dedicated
attributes array property for any applications using this library to
track state on nodes.
(This isn't something I personally need, though. I haven't found any use cases in applications I've used before, largely because static analyzers I'm aware of use php-parser or php-ast for analysis.)

- https://github.com/microsoft/tolerant-php-parser/network/dependents?dependent_type=PACKAGE would be a place to start looking. php-language-server is not installable in php 8.x right now for unrelated reasons.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -2128,6 +2128,7 @@ class Parser {
             }
 
             if ($shouldOperatorTakePrecedenceOverUnary) {
+                /** @var UnaryOpExpression $unaryExpression */
                 $unaryExpression = $leftOperand;
                 $leftOperand = $unaryExpression->operand;
             }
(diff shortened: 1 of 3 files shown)


roblourens (approved):
Thanks!

PR #364, opened 2022-01-01
callableExpression can also be a QualifiedName...

dantleech:
...when the callable expression is a function


src/Node/Expression/CallExpression.php
```diff
@@ -8,10 +8,11 @@
 
 use Microsoft\PhpParser\Node\DelimitedList;
 use Microsoft\PhpParser\Node\Expression;
+use Microsoft\PhpParser\Node\QualifiedName;
 use Microsoft\PhpParser\Token;
 
 class CallExpression extends Expression {
-    /** @var Expression */
+    /** @var QualfifiedName|Expression */
```
roblourens, line 15: Typo :)

dantleech: oops updated

PR #368, opened 2022-08-09
Support php 8.2 readonly classes, constants in traits, `true` type

TysonAndre:
1. Support https://wiki.php.net/rfc/readonly_classes - the class modifiers become repeatable, it becomes a compile error rather than a parse error to repeat modifiers or to combine abstract&final
2. Support https://wiki.php.net/rfc/constants_in_traits
3. Support parsing https://wiki.php.net/rfc/true-type


TysonAndre:
@roblourens thoughts?

roblourens (approved):
Thanks!

PR #370, opened 2022-08-11
Set up github workflows for multiple php versions

TysonAndre:
Run this alongside Travis CI

- Start testing this with php 8.2, the feature freeze was in July with the first beta https://wiki.php.net/todo/php82
- php betas are often late for getting stable tags in travis - https://docs.travis-ci.com/user/languages/php/#php-versions does not have images for the php 8.2 nightly (the `master` and `nightly` tags will eventually point to 8.3 and there might not be an 8.2 for a while after that)
- Start with just phpunit tests for now, to make this easier to review
- This will give faster feedback if there are travis queueing delays, and show results for individual php versions separately
- Running dockerfiles makes it easier to develop/run the same checks locally (if there is a build failure) or to customize this (e.g. dependency versions) later

See https://github.com/TysonAndre/tolerant-php-parser/pull/2 for results

Note that for this to run in this repo, github workflows must first be enabled by the maintainers (e.g. by merging this or by starting the process of setting up github actions)


roblourens (commented):
This looks great, thanks. I'd be happy to leave Travis entirely. Could you just add .sh file extensions to the script files?

TysonAndre:
> This looks great, thanks. I'd be happy to leave Travis entirely. Could you just add .sh file extensions to the script files?

I added the `.sh` extensions, this can be extended once checks are merged and running

roblourens (approved):

PR #374, opened 2022-08-18
Support php 'never' type, change parsing of 'void'

### LINKED ISSUE #372: Support php 8.1 `never` return type, add to returnTypeDeclarationTokens
https://wiki.php.net/rfc/noreturn_type#vote (secondary vote was for `never`)

### LINKED ISSUE #373: Using `void` as a parameter/property type is a compilation error, not a parse error
```php
php > function foo(void $x){ }

Fatal error: void cannot be used as a parameter type in php shell code on line 1
```

Using void in parameter names and php 7.4 property types parses, but causes a CompileError

However, `static` is a parse error in parameter types

Low priority since valid code wouldn't do that

TysonAndre:
Similar to 'void', it's a compile time error to use void as a
parameter/property type, not a parse time error.
But `static` is a parse error.
(It can also be used in union types)

Remove redundant tokens added to merged returnTypeDeclarationTokens

Closes #373
Closes #372


roblourens:
There are no test cases that have to be changed as a result of the 'void' change? Should there be?

TysonAndre:
I added a test that void/never are now allowed in parameter types. There weren't any before.

Aside: tests/cases/parser/staticReturnType2.php already tests that `static` is a parse error in parameter types (`')' expected`)

roblourens (approved):
Thanks!

PR #380, opened 2022-08-26
Speed up running the framework validation test suite

TysonAndre:
1. We're parsing thousands of files and there are many tokens per file.
   The overhead of all the assert statements is noticeable.
   Optimizing out the check (when they'd pass) speeds this up 3x from 90 seconds to 30 seconds - notably, function calls in php are slow, and phpunit has multiple levels of them
2. Don't write files only to unlink them later.
   Instead, save the failed file only on test suite failure


TysonAndre:
@roblourens - thoughts? The speedup for this would be convenient for local development, and #382  would make the validation test suite pass so that it can be required in CI again

roblourens (approved):
Great speedup! PHP... 😆

PR #385, opened 2022-09-22
PHPStan Level 3

dantleech:
:wave: in [Phpactor](https://github.com/phpactor/phpactor) one of the most common types of error is accessing a member on NULL on TP parser nodes. So I thought I'd try and increase the type coverage.

This PR:

- Updates PHPStan to level 3.
- Fixes type hints (avoiding refactoring code).
- Adds PHPStan as a dev dependency (it ships as a PHAR now, so no conflicts).
- Removes the overridden public `opeand` property on the children of `UnaryExpression`

As a follow up question is it worth ditching Travis and adding PHPStan to the `main` GHA workflow?

We should be able to get to PHPStan level 5 pretty easy (30 errors) but level 6 is 368 errors :sweat_smile:

PR #386, opened 2022-09-25
Move validation test suite to GitHub Workflows

TysonAndre:
- Perform a shallow clone of submodules analyzed in validation test suite in GitHub workflows (should take around 30 seconds)
- Switch badge in README.md to GitHub Workflows and update default branch for badge.
- Remove travis config
- After #385 is merged run_phpstan.sh can remove the separate install step (as a followup PR)


TysonAndre:
This is ready for review, no other changes are planned

- phpstan successfully ran in php 7.2 (needed to use matrix.VALIDATION)
- The validation folder is now mounted as a read-only volume when running tests with ci/run_tests_dockerized.sh and the test suite runs on the 10000+ php files found in validation/frameworks
- This was rebased against the latest commits

roblourens (dismissed):
Awesome, thank you for doing this!

rzhao271 (dismissed):

ci/run_phpstan.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# -x Exit immediately if any command fails
+# -e Echo all commands being executed.
+# -u Fail for undefined variables
+set -xeu
+# This installs and runs phpstan
+# TODO: Remove separate phpstan install step after https://github.com/microsoft/tolerant-php-parser/pull/385 is merged
```
roblourens, line 7: I've merged 385 and removed a merge conflict with main, do you want to remove this now?

TysonAndre:
I undid the changes to php versions in github actions to avoid conflicts with https://github.com/microsoft/tolerant-php-parser/pull/406/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3 and updated composer.json again

Unit tests and validation steps aren't run until the workflow run is approved https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

ci/run_phpstan.sh
roblourens: This separate step can be removed now though right?

roblourens:
It looks like phpstan ran out of memory

```
     PHPStan process crashed because it reached configured PHP memory limit:   
     128M      
```

dantleech:
> It looks like phpstan ran out of memory

if that's the only issue it can be fixed with `phpstan --memory-limit=4G` or similar cc @TysonAndre

Issue #381, opened 2022-08-26
Support `__halt_compiler` as a member of the top level statements


TysonAndre:
This would allow validation test suites to pass again

- On the top level scope, emit a new HaltCompilerStatement with haltCompilerToken, openParen, closeParen, semicolonToken, and data of type T_INLINE_HTML
- It's a compile error on an inner statement, so not really important to parse it, but that would be with a MissingToken for `data` if it was parsed in an inner statement.

```
// Zend/zend_language_parser.y
top_statement:
//  ....
	|	T_HALT_COMPILER '(' ')' ';'
			{ $$ = zend_ast_create(ZEND_AST_HALT_COMPILER,
			      zend_ast_create_zval_from_long(zend_get_scanned_file_offset()));
			  zend_stop_lexing(); }
```

```
php > var_export(token_get_all('<?php echo 2; __halt_compiler(); ?><?php fake data'));
array (
  0 => 
  array (
    0 => 382,
    1 => '<?php ',
    2 => 1,
  ),
  1 => 
  array (
    0 => 324,
    1 => 'echo',
    2 => 1,
  ),
  2 => 
  array (
    0 => 385,
    1 => ' ',
    2 => 1,
  ),
  3 => 
  array (
    0 => 309,
    1 => '2',
    2 => 1,
  ),
  4 => ';',
  5 => 
  array (
    0 => 385,
    1 => ' ',
    2 => 1,
  ),
  6 => 
  array (
    0 => 363,
    1 => '__halt_compiler',
    2 => 1,
  ),
  7 => '(',
  8 => ')',
  9 => ';',
  10 => 
  array (
    0 => 313,
    1 => ' ?><?php fake data',
    2 => 1,
  ),
)
php > echo token_name(313);
T_INLINE_HTML
```

PR #389, opened 2022-09-25
Fix handling of HaltCompilerStatement

TysonAndre:
Make it easier to calculate `__COMPILER_HALT_OFFSET__` and make properties other than `data` non-null.
Combine all tokens after halt compiler statement into an InlineHtml token that goes into HaltCompilerStatement->data if there are bytes.
Related to #381


TysonAndre:
This is ready for review. The ParserInvariantTest asserts that the tokens combine into the original file for the test cases of `__halt_compiler()`

TysonAndre:
One last change: A real InlineHtml node would have no leading whitespace/comments, so make fullStart = start.
- In places that aren't HaltCompilerStatement (EchoStatement), it's always at the start of a file or after `?>`/`?>\n`, and the newline is part of the close tag token instead of the InlineHtml token

tests/cases/parser/haltCompiler4.php.tree
```diff
@@ -27,58 +27,16 @@
...
                         "error": "MissingToken",
                         "kind": "SemicolonToken",
                         "textLength": 0
                     },
-                    "data": null
-                }
-            },
-            {
-                "error": "SkippedToken",
-                "kind": "ColonColonToken",
```
TysonAndre, line 40: The reasoning in updating the tests I'd added in the previous PR is that HaltCompiler should halt the compiler and not parse any more statements, so I'd treat the snippet like `__halt_compiler[();]::foo();` with inline html of `::foo();\n`
(this token can't be used in other statements/expressions and is a syntax error)

```
<?php
// invalid
__halt_compiler::foo();
```

The MissingToken in the original tree were due to token_get_all() returning at most 3 tokens after T_HALT_COMPILER then returning T_INLINE_HTML for the rest

roblourens (approved):

Issue #390, opened 2022-10-03
Regression with namespace names vs. MissingToken


dantleech:
See https://github.com/microsoft/tolerant-php-parser/pull/379

---

I think the above PR causes a regression when calling `$node->getNamespacedName()` where the namespace has a `MissingToken` for `name`:

```
Fatal error: Uncaught TypeError: Microsoft\PhpParser\ResolvedName::buildName(): Argument #1 ($tokens) must be of type array, nul
l given, called in /home/daniel/www/phpactor/phpactor/vendor/microsoft/tolerant-php-parser/src/NamespacedNameTrait.php on line 35 and defined in /home/daniel/www/phpactor/phpactor/vendor/microsoft/tolerant-php-parser/src/ResolvedName.php:15
Stack trace:                                                                                                                    #0 /home/daniel/www/phpactor/phpactor/vendor/microsoft/tolerant-php-parser/src/NamespacedNameTrait.php(35): Microsoft\PhpParser\
ResolvedName::buildName()
```

i.e. 

```
        if ($namespaceDefinition->name !== null) {
            $resolvedName = ResolvedName::buildName($namespaceDefinition->name->nameParts, $content);
```

now tries to access `nameParts` on NULL, there may be other places (for info there are also at some parts of Phpactor that were depending on this value so would at least be a B/C break)

zobo:
Got bitten by this too. https://github.com/zobo/php-language-server/pull/47

dantleech:
Looks like we fixed it in #391 so I'll close this now

PR #391, opened 2022-10-03
Fix regression with getNamespacedName()

dantleech:
See #390  

This code depended on `name` being NULL but now it seems to be MissingToken cc @TysonAndre


TysonAndre (approved):
lgtm - the phpdoc comment on the property src/Node/Statement/NamespaceDefinition.php could also be changed to `/** @var QualifiedName|MissingToken|null */` to avoid this when parsing `namespace ;`

roblourens (approved):

PR #395, opened 2022-12-21
fix: interface properties

nataneb32:
interfaces property declaration as broken

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -3516,10 +3516,7 @@ class Parser {
                     return $this->parseAttributeStatement($parentNode);
 
                 default:
-                    $missingInterfaceMemberDeclaration = new MissingMemberDeclaration();
-                    $missingInterfaceMemberDeclaration->parent = $parentNode;
-                    $missingInterfaceMemberDeclaration->modifiers = $modifiers;
-                    return $missingInterfaceMemberDeclaration;
+                    return $this->parseRemainingPropertyDeclarationOrMissingMemberDeclaration($parentNode, $modifiers);
             }
         };
     }


src/Parser.php
```diff
@@ -3516,10 +3516,7 @@ private function parseInterfaceElementFn() {
                     return $this->parseAttributeStatement($parentNode);
 
                 default:
-                    $missingInterfaceMemberDeclaration = new MissingMemberDeclaration();
-                    $missingInterfaceMemberDeclaration->parent = $parentNode;
-                    $missingInterfaceMemberDeclaration->modifiers = $modifiers;
-                    return $missingInterfaceMemberDeclaration;
+                    return $this->parseRemainingPropertyDeclarationOrMissingMemberDeclaration($parentNode, $modifiers);
```
dantleech, line 3519: possible to add a test for this? or what is a code example that would invoke this change?

TysonAndre:
This change has a point, though it seems unrelated to the root cause of the bug in the linked ticket. PHP internally parses `class_statement_list` the same way for classes, enums, traits, interfaces. So property declarations in interfaces would be a **compile-time error**, not a parser error, and tolerant-php-parser is meant to be a tolerant parser (and applications using this should be checking for compilation errors)

https://github.com/php/php-src/blob/PHP-8.3.1/Zend/zend_language_parser.y#L622-L638

TysonAndre:
I guess `.php` stubs for UnitEnum/BackedEnum interfaces in php would have fake $name/$value(BackedEnum) properties for documentation in IDEs/editors?

This is still a compile-time error in php 8.3 (not a parser error). My personal preference is to unify the parsing of all the class-likes into one helper method but I hadn't found it that urgent

```
php > interface foo { public int $x; }

Fatal error: Interfaces may not include properties in php shell code on line 1
```

PR #396, opened 2023-01-05
Phpstan level 4

### LINKED ISSUE #397: Quoted string parsing doesn't work as expected
As requested in this [PR](https://github.com/microsoft/tolerant-php-parser/pull/396#discussion_r1063150728)

Currently when parsing

```
<?php "'hello world'"';
```

Results in:

```
hello world
```

and not

```
'hello world'
```

same for single and double quotes.

dantleech:
This PR:

- adds support for PHPStan Level 4
- fixes string literal parsing and adds a test for it  (fixes #397)

### DIFF
--- a/src/Node.php
+++ b/src/Node.php
@@ -587,7 +587,7 @@ abstract class Node implements \JsonSerializable {
             ? $this
             : $this->getFirstAncestor(NamespaceDefinition::class, SourceFileNode::class);
 
-        /** @phpstan-ignore-next-line TODO: can this happen? test with framework test cases */
+        /** @phpstan-ignore-next-line result is always false -- not sure this can happen */
         if ($namespaceDefinition instanceof NamespaceDefinition && !($namespaceDefinition->parent instanceof SourceFileNode)) {
             $namespaceDefinition = $namespaceDefinition->getFirstAncestor(SourceFileNode::class);
         }
--- a/src/Node.php
+++ b/src/Node.php
@@ -437,10 +437,7 @@ abstract class Node implements \JsonSerializable {
     public function getDescendantNodeAtPosition(int $pos) {
         foreach ($this->getChildNodes() as $child) {
             if ($child->containsPosition($pos)) {
-                $node = $child->getDescendantNodeAtPosition($pos);
-                if (!is_null($node)) {
-                    return $node;
-                }
+                return $child->getDescendantNodeAtPosition($pos);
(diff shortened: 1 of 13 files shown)


src/Node/StringLiteral.php
```diff
@@ -25,17 +25,29 @@ class StringLiteral extends Expression {
...
+                if ($startQuote === '\'') {
+                    return rtrim(substr($value, 1), '\'');
+                }
+
+                if ($startQuote === '"') {
+                    return rtrim(substr($value, 1), '"');
+                }
+            }
+
+            return '';
```
dantleech, line 50: this is a feature/bug fix

roblourens: If there isn't an issue for this, could you file one and link this PR to it?

dantleech: sure: https://github.com/microsoft/tolerant-php-parser/issues/397

src/Node.php
```diff
@@ -590,6 +587,7 @@ public function getNamespaceDefinition() {
             ? $this
             : $this->getFirstAncestor(NamespaceDefinition::class, SourceFileNode::class);
 
+        /** @phpstan-ignore-next-line TODO: can this happen? test with framework test cases */
         if ($namespaceDefinition instanceof NamespaceDefinition && !($namespaceDefinition->parent instanceof SourceFileNode)) {
```
dantleech, line 591: should either ignore this or remove the ignore and update the code and hope that it does never happen.

tests/unit/Node/StringLiteralTest.php
```diff
@@ -0,0 +1,56 @@
+<?php
```
roblourens, line 1: Would it make sense to use the pattern in `tests/cases/parser` for this?

dantleech: hm, in what sense? 

keeping this test case but globbing over `../cases/parser/stringLitreal.php`  or something else?

unless I'm missing something there is no way to assert anything with the parser tests as they check the AST structure, while this test is testing a method on the Node.

roblourens: Sorry, I didn't look at it closely enough, yes this makes sense as-is.

roblourens (approved):

roblourens:
Thanks!

PR #398, opened 2023-04-20
Support Enum Interface clause

dantleech:
This PR adds the missing support for:

```
enum Foo: string implements Bar
{
}
```

as previously the intreface clause was not parsed.


roblourens (approved):
Thank you!

DonJayamanne (approved):

Issue #400, opened 2023-05-15
Why is QualifiedName::getResolvedName disabled for Node\TraitUseClause?


zobo:
When I need to resolve the FQN of the `use TestTrait` I cannot use the handy getResolvedName. Why is that?

```php
namespace YYY;

use \XXX\TestTrait;

class TestClazz
{
	use TestTrait;

```

https://github.com/microsoft/tolerant-php-parser/blame/84ce745a39619b10734c8e6b736e055b9fa64d13/src/Node/QualifiedName.php#L88

Thanks!

roblourens:
I don't know- I was just looking at the test cases and I'm not sure why you can use it on some code that references that name, but not in the `use` statement. If it makes sense, we could change that.

zobo:
Hi, thanks for the feedback. Maybe it has something to do with the crazy things you can do with the `insteadof` and `as` keywords. See "Conflict resolution" in https://www.php.net/manual/en/language.oop5.traits.php .

zobo:
But then again, I can't imagine this be the reason... I tried commenting out `Node\TraitUseClause` to see if any of the tests in php-intellisense break and there was no problem - if something, some things I missed got fixed.


```php
    public function getResolvedName($namespaceDefinition = null) {
        // Name resolution not applicable to constructs that define symbol names or aliases.
        if (($this->parent instanceof Node\Statement\NamespaceDefinition && $this->parent->name->getStartPosition() === $this->getStartPosition()) ||
            $this->parent instanceof Node\Statement\NamespaceUseDeclaration ||
            $this->parent instanceof Node\NamespaceUseClause ||
            $this->parent instanceof Node\NamespaceUseGroupClause ||
            /*$this->parent->parent instanceof Node\TraitUseClause ||*/
            $this->parent instanceof Node\TraitSelectOrAliasClause
        ) {
            return null;
        }
```

I have not yet managed to wrap my head around `TraitSelectOrAliasClause` where in the parsers it's being used.

PR #401, opened 2023-05-16
fix: Allow getResolvedName to resolve QualifiedName in TraitUseClause

zobo:
Discussion in https://github.com/microsoft/tolerant-php-parser/issues/400


roblourens (commented):
Thanks for the PR. The change is fine to me, but there are tests for this, so please update those too.

roblourens (commented):
Thanks for the PR. The change is fine to me, but there are tests for this, so please update those too.

zobo:
Ah, managed to miss those due to all the noise running phpunit on windows. Will look at those asap.

zobo:
@microsoft-github-policy-service agree

roblourens (approved):

DonJayamanne (approved):

PR #410, opened 2024-08-11
Fix php 8.4 notices about implicitly nullable parameters

TysonAndre:
Nullable types were supported since php 7.1, this library requires php >= 7.2, https://wiki.php.net/rfc/deprecate-implicitly-nullable-types deprecates this in php 8.4

And fix notice about deprecated string encapsulation syntax


roblourens (approved):
Thanks!

lszomoru (approved):

shaismithu (commented):
how to become pro user

shaismithu:
hello how its work can you explain

PR #411, opened 2024-08-11
Support parsing of PHP 8.4 property hooks

TysonAndre:
As part of this, stricten the parsing of property names and use a narrower syntax but reuse AssignmentExpression for backwards compatibility of applications that originally used tolerant-php-parser's result of parseExpression.

References:
https://wiki.php.net/rfc/property-hooks and the linked php-src PR's zend_language_parser.y

https://wiki.php.net/rfc/property-hooks#abbreviated_syntax https://wiki.php.net/rfc/property-hooks#interaction_with_constructor_property_promotion


src/Parser.php
```diff
@@ -3423,12 +3429,115 @@ private function parsePropertyDeclaration($parentNode, $modifiers, $questionToke
...
+        return function ($token) {
+            return \in_array($token->kind, self::PROPERTY_HOOK_START_TOKENS, true);
+        };
+    }
+
+    /**
+     * @param PropertyHooks $parentNode
+     * @return PropertyHook
+     */
+    private function parsePropertyHook($parentNode) {
```
dantleech, line 3512: we can add a type hint here I think?

src/Node/PropertyHook.php
```diff
@@ -0,0 +1,44 @@
...
+    /** @var DelimitedList\ParameterDeclarationList|null */
+    public $parameters;
+    /** @var Token|null */
+    public $closeParen;
+    /** @var Token|null */
+    public $arrowToken;
+    /** @var CompoundStatementNode|Expression|Token|null */
+    public $body;
+    /** @var Token|null */
+    public $semicolon;
```
dantleech, line 31: not related to this PR but I guess we could consider bumping the minimum version to 8.0 or even 7.4?

dantleech:
I've borrowed this work and made some additional changes in Phpactor's fork here: here: https://github.com/phpactor/tolerant-php-parser/pull/2

PR #412, opened 2024-09-03
Support parsing php 8.3 typed class constants

dantleech:
https://wiki.php.net/rfc/typed_class_constants#inheritance_and_variance As of php 8.3, class constants can have the same types as parameters (including union, intersection, and DNF types)

Start testing with php 8.3


dantleech:
should be fixed now

PR #415, opened 2025-01-13
PHP8.4: Support new without parenthesis

dantleech:
(from Phpactor's fork - not tested extensively)

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -3047,7 +3047,8 @@ class Parser {
             $expression instanceof SubscriptExpression ||
             $expression instanceof ScopedPropertyAccessExpression ||
             $expression instanceof StringLiteral ||
-            $expression instanceof ArrayCreationExpression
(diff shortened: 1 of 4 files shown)


src/Parser.php
```diff
@@ -3289,6 +3290,11 @@ private function parseObjectCreationExpression($parentNode) {
             $objectCreationExpression->classMembers = $this->parseClassMembers($objectCreationExpression);
         }
 
+        // PHP8.4 new with no parenthesis
+        if ($this->getCurrentToken()->kind === TokenKind::ArrowToken) {
+            return $this->parsePostfixExpressionRest($objectCreationExpression);
```
dantleech, line 3295: In line with the rest of this libaray this should probably have a run time `php_version() > 80400`? 

(incidentally would be good to refactor so that it didn't depend on the runtime PHP version)

zobo: +1 for being able to configure at what PHP level the parser works

